PR #10, opened 2026-02-20
security(oauth): harden token handling to ensure OAuth stays in Anthr…

johnzfitch:
The key design principle driving all these changes is token non-interference: our compatibility layer should demonstrably never capture, store, or relay OAuth tokens. The unmodified Claude Desktop renderer and Claude Code CLI each handle their own authentication directly with Anthropic's servers. Our stubs exist purely to satisfy the Electron IPC contract (they must respond to the channel), but they should be provably inert regarding credentials.

This compatibility layer ensures that this repository never captures, stores, or forwards OAuth tokens. The key architectural guarantee these changes enforce is structural impossibility: it's not just that our code doesn't forward tokens — it can't. The allowlist pattern means that even if Anthropic adds new token-bearing env vars in future Claude Desktop updates, they'd be automatically excluded unless explicitly added to the allowlist. 

Changes:
- P0: addApprovedOauthToken is a documented no-op (_token discarded)
- P0: sdk_bridge.js uses env allowlist instead of spreading process.env
- P1: filterEnv blocks credential-like additionalEnv keys (regex gate)
- P1: AuthRequest documents it never intercepts OAuth callbacks
- P2: Auth_$_doAuthInBrowser validates URL origin (Anthropic domains only)
- Add OAUTH-COMPLIANCE.md with full audit of every auth-related location

https://claude.ai/code/session_01XLqRFKFca9sMaxDu95meSt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,14 @@ claude-launcher-fixed.sh
 # Internal docs (may contain paths, session IDs, etc.)
 docs/plans/
 docs/logs/
+docs/ARCHITECTURE.md
+ARCHITECTURE.md
+auth_code_reference.txt
+auth_findings_summary.md
+
+# Local Claude Code config (session notes, local settings)
+.claude/session-resume.md
+.claude/settings.local.json
 
 # OS files
 .DS_Store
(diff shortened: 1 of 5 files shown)


johnzfitch:
@claude full code review

PR #46, opened 2026-03-13
Fix Linux cowork plugin sync and remote chat renderer regressions

johnzfitch:
## Summary
- hydrate cowork plugin state and materialize installed plugin caches for Linux sessions
- restore live local session assistant/thinking updates without filtering incremental events
- remove the broad renderer API JSON fallback shim that was mutating claude.ai chat/artifact responses

## Verification
- rebuilt local app.asar after the wrapper/stub changes
- checked latest log at /home/zack/errors.txt for successful completion requests and restored quick access overlay calls
- noted one residual `message_store_sync_blocked` event remains in the latest log for follow-up

### DIFF
--- a/stubs/frame-fix/frame-fix-wrapper.js
+++ b/stubs/frame-fix/frame-fix-wrapper.js
@@ -673,8 +673,135 @@ function cloneMessageContent(content) {
     if (!block || typeof block !== 'object') {
       return block;
     }
-    return { ...block };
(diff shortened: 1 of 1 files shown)


johnzfitch:
Superseded by v4.0.0 release. Plugin sync and renderer regression fixes are now handled by the new stubs/cowork/ orchestration layer.

PR #72, opened 2026-04-05
fix: compatibility with asar v1.569.0 (systemPreferences, native stub paths, bridge dispatch)

wlcarden:
## Summary

Fixes four issues that prevent launch and break dispatch on asar v1.569.0:

- **Launch crash**: asar calls `systemPreferences.setUserDefault()` at top-level module eval — a macOS-only API that doesn't exist on Linux. Added stubs for `setUserDefault`, `getUserDefault`, `removeUserDefault`, `promptTouchID`, and `canPromptTouchID`.
- **Native stub silent failure**: `require('../../cowork/dirs.js')` resolves correctly from `stubs/@ant/claude-native/` (2 levels) but breaks from `node_modules/@ant/claude-native/` (3 levels). Fixed to `../../../cowork/`. Also added native stub sync to `launch.sh` alongside the existing swift stub sync.
- **Bridge registration skipped**: the asar's `cliPluginBridge` initializer checks for `setGuestRequestCallback` and `sendGuestResponse` on the VM API. Without them, bridge registration is skipped. Added both methods to the swift stub.
- **Dispatch "can't connect"**: 12 bridge handler overrides (`kickBridgePoll`, `resetBridge`, `sessionsBridgeStatus`, etc.) were intercepting the asar's own `LocalAgentModeSessionManager` handlers and replacing them with no-ops. This prevented the bridge WebSocket (`wss://bridge.claudeusercontent.com`) from ever connecting. Removed the overrides so the asar's handlers run unimpeded.

## Security notes

Per CONTRIBUTING.md, changes to security-sensitive files:

- **`stubs/@ant/claude-swift/js/index.js`**: `setGuestRequestCallback` stores a callback reference on the instance. `sendGuestResponse` is a no-op trace log. Neither method handles credentials or spawns processes.
- **`stubs/@ant/claude-native/index.js`**: require path fix only — no behavioral change.

## Test plan

- [x] All 354 tests pass (`node --test tests/node/current-path/*.test.cjs`)
- [x] App launches without crash on Linux (previously crashed on `setUserDefault`)
- [x] `[claude-native-stub] claude-native stub loaded successfully` (previously `Failed to load Claude Native`)
- [x] `[cliPluginBridge] registered` (previously `VMAPI missing guest-request methods`)
- [x] Dispatch connects and is usable (previously showed "can't connect")

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -23,12 +23,18 @@ fi
 
 STUB_FILE="linux-app-extracted/node_modules/@ant/claude-swift/js/index.js"
 STUB_SRC_FILE="stubs/@ant/claude-swift/js/index.js"
+NATIVE_STUB_FILE="linux-app-extracted/node_modules/@ant/claude-native/index.js"
+NATIVE_STUB_SRC="stubs/@ant/claude-native/index.js"
 
 # Ensure the extracted app tree has the latest stubs baked in before packing.
 # This avoids relying on runtime module interception (ESM import() bypasses Module._load).
 if [ -f "$STUB_SRC_FILE" ]; then
   mkdir -p "$(dirname "$STUB_FILE")"
   cp -f "$STUB_SRC_FILE" "$STUB_FILE"
+fi
+if [ -f "$NATIVE_STUB_SRC" ]; then
+  mkdir -p "$(dirname "$NATIVE_STUB_FILE")"
+  cp -f "$NATIVE_STUB_SRC" "$NATIVE_STUB_FILE"
 fi
 
 # Sync frame-fix files so wrapper changes take effect without a full reinstall
--- a/stubs/@ant/claude-native/index.js
+++ b/stubs/@ant/claude-native/index.js
@@ -22,8 +22,8 @@ const { ipcMain } = require('electron');
 const EventEmitter = require('events');
 const path = require('path');
 const fs = require('fs');
-const { createDirs } = require('../../cowork/dirs.js');
-const { redactCredentials } = require('../../cowork/credential_classifier.js');
+const { createDirs } = require('../../../cowork/dirs.js');
+const { redactCredentials } = require('../../../cowork/credential_classifier.js');
 
 const LOG_PREFIX = '[claude-native-stub]';
 const DIRS = global.__coworkDirs || createDirs();
--- a/stubs/@ant/claude-swift/js/index.js
+++ b/stubs/@ant/claude-swift/js/index.js
@@ -393,10 +393,17 @@ function createMountSymlinks(sessionName, additionalMounts) {
     }
 
     // Construct the full host path
-    // Empty path means homedir itself
+    // The asar provides paths relative to the VM root (e.g., "home/user/.config/...").
+    // Prepending homedir doubles the path. Detect this: if the relative path starts
+    // with a known top-level directory (home/, root/, tmp/, etc.), treat it as an
+    // absolute path with a missing leading slash rather than joining with homedir.
+    const asAbsolute = '/' + relativePath;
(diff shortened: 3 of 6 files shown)


johnzfitch:
Gonna leave this open for now. Remote/dispatch is disabled atm due to security problems. Going to consider the current stubs as protective for now.

If you want to resubmit, add just the crash fixes for the latest version. 

It's important that it doesn't break backwards compatibility for those who haven't updated yet so add the new classifiers with comment markers to remove the old with the current date.

I am going to keep the remote/dispatch surfaces unimplemented for now due to safety concerns w/ both surfaces.

johnzfitch:
Thanks for the detailed writeup @wlcarden — the bridge-dispatch analysis specifically saved us real time on triaging #77.

**Three of the four issues you identified have already landed through other merges**, so there's less left in this PR than it looks:

- `systemPreferences.setUserDefault` / `getUserDefault` stubs → merged via #67 (`cab72fa`).
- `@ant/claude-native` require path + native-stub sync on launch → merged via #68 (`eb8ff47`) + `64864a5`.
- The doubled-homedir mount-symlink fix → merged via #74 (`4047f81`), shipping in v4.3.1.

**The bridge dispatch piece is the one we're deliberately passing on** — this is an intentional call, not a "needs rework" signal.

The Linux port keeps `wss://bridge.claudeusercontent.com` disabled by design. The bridge transport carries server-initiated messages that drive local handlers (`respondBridgePermissionPreflight`, `onBridgePermissionPreflight`, bridge state mutation, etc.), and those code paths haven't been exercised or audited against the Linux-port threat model — it's a macOS-VM-first feature upstream. PR #89 (v4.3.2 prep) reinforces this by setting `getBridgeConsent → {consented: false}` explicitly.

The downside is the "can't connect" UX you correctly identified as affecting #77. The plan there is to change the stubs from "pretend to succeed" to returning a clean `{supported: false, reason: 'dispatch not available on Linux port'}` so the UI can render an unsupported-state message instead of a failed connection.

Going to close this as superseded (overlapping hunks) / won't-fix (for the bridge portion). Really appreciate the clean commit history and the root-cause writeup — please keep filing issues and PRs, this one was high-signal even if we landed most of it through a different path.

---

johnzfitch:
Gonna leave this open for a bit for others to weigh in. I personally think the better security standard is off by default, easy to fork and wire if you want to make that decision for yourself. Implement something form fit for your existing security architecture rather than implementing a one size fits all approach that even Anthropic has problems handling on a closed-system Mac. Your novel architecture will be much more secure than anything I put in place at a project level. I want to try to make sure that high-risk surfaces such as dispatch are handled really thoughtfully and carefully. But I am open to hearing what you all think and am just approaching this out of an abundance of caution.

johnzfitch:
Closing in favor of #102 which incorporates your contributions with additional hardening.

Your fixes that landed:
- `systemPreferences.setUserDefault` / `getUserDefault` / `removeUserDefault` / `promptTouchID` / `canPromptTouchID` stubs
- `claude-native/index.js` require path fix (`../../` &rarr; `../../../`)
- `launch.sh` native stub sync
- `setGuestRequestCallback` / `sendGuestResponse` on the VM API

The `looksAbsolute` path heuristic was replaced with a tighter `startsWith(homedir + '/')` check to prevent prefix collisions (e.g. `/home/user2/` matching `/home/user`). Bridge handler overrides were kept as no-ops rather than removed &mdash; this is an intentional design choice for the Linux port's network posture.

Thank you @wlcarden for the thorough investigation and fix &mdash; the systemPreferences crash and guest request methods were critical for v1.569.0 compatibility.

---

Issue #82, opened 2026-04-18
[Bug] Claude Code 2.1.114 breaks cowork — `vm.spawn` rejects `bash`


janithcooray:
### Area

Cowork session (spawn / spawn fails)

### What happened?

After updating Claude Code to 2.1.109, cowork fails on any tool call that needs a shell. The security allowlist in @ant/claude-swift blocks bash as an unrecognized command, because current Claude Code versions invoke shell-tool operations via bash -c "cd … && cmd" instead of calling the resolved claude binary directly. No spawn ever reaches the OS, so the failure surfaces to the UI as a crash (exit-126-style: "command invoked cannot execute").
Cowork worked fine until Claude Code was updated. Only ran ./install.sh, no other changes.

## TEMP FIX
revert back to 2.1.109

### Steps to reproduce

1. Had a working cowork setup on Claude Code 2.1.109.
2. Ran `./install.sh` — it auto-updated Claude Code to 2.1.114.
3. Cowork now crashes on any tool call that needs a shell.

### Relevant log output

```shell
Claude Code 2.1.114 invokes shell tool calls as `bash -c "cd … && cmd"` instead of calling the `claude` binary directly. The allowlist in `@ant/claude-swift` rejects `bash` as an unvetted command, so the spawn never runs.

From `~/.local/state/claude-cowork/logs/claude-swift-trace.log`:

\`\`\`
SECURITY: Unexpected command blocked: "bash" (type=string)
=== VM.SPAWN CALLED ===
vm.spawn() command=bash
vm.spawn() args=["-c","cd /sessions/<session>/mnt/<project>/backend && go build ./... 2>&1"]
\`\`\`

2.1.109 does not do this; it calls the claude binary directly, which is on the allowlist.
```

### Distro and desktop environment

Gentoo / KDE

### claude-cowork-linux version / commit

fcd8111

### Claude Desktop version (app.asar)

2.1.114

### Checklist

- [x] I ran `./install.sh --doctor` and included its output if relevant
- [x] I checked the log file and redacted any tokens before pasting
- [x] I searched existing issues and this isn't a duplicate

janithcooray:
patched with #84

PR #84, opened 2026-04-18
Allow bash -c invocations in vm.spawn allowlist

janithcooray:
Recent Claude Desktop builds issue shell tool calls as `bash -c \"cd <dir> && <cmd>\" `

instead of invoking the claude binary directly. The allowlist in session_orchestrator.js rejects bare 'bash' as an unvetted command, blocking every shell tool call with 'SECURITY: Unexpected command blocked' and surfacing in the UI as an immediate session crash.

Add a branch that resolves bash to /usr/bin/bash or /bin/bash (both already covered by the existing allowedPrefixes list), so the spawn proceeds through the normal path with the same security posture as any other /usr/bin binary.

## What does this change?

Bypassing bash `SECURITY: Unexpected command blocked: "bash" (type=string)` on latest co work v2.1.114

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

- Distro / desktop: Gentoo / KDE - Wayland
- Tested with `./install.sh --doctor`: Yes
- Tested a full Cowork session: Yes

## Security impact

spawn only: resolved a new hostCommand and let the existing spawn path run unchanged

The patch only adds a branch that resolves `bash` / basename `bash` to `/usr/bin/bash` or `/bin/bash` — both already covered by `allowedPrefixes`. The dispatch then falls through to the existing spawn path, inheriting all env filtering, credential classification, and log redaction unchanged.

- [x] This change does not touch credential handling, token passthrough, or process spawning

## Checklist

- [x] `./install.sh --doctor` passes cleanly on my system

### DIFF
--- a/stubs/cowork/session_orchestrator.js
+++ b/stubs/cowork/session_orchestrator.js
@@ -522,6 +522,20 @@ class SessionOrchestrator {
     ) {
       hostCommand = resolveClaudeBinaryPath();
       trace('Translated command: ' + normalizedCommand + ' -> ' + hostCommand);
+      } else if (
+      normalizedCommand === 'bash' ||
+      commandBasename === 'bash'
+    ) {
+      const bashCandidates = ['/usr/bin/bash', '/bin/bash'];
+      hostCommand = bashCandidates.find((c) => fs.existsSync(c));
+      if (!hostCommand) {
+        trace('SECURITY: bash requested but not found on host');
+        if (typeof onError === 'function') {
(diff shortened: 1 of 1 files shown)


janithcooray:
to patch #82

F1nny:
Was doing the same and saw this :laughing: so closed the dupe, much better especially /bin vs /usr/bin - only rec would be adding sh as well, thanks!

PR #75, opened 2026-04-09
Security enhancements (nonalarmist title lol)

johnzfitch:
Deny-by-default for all permission gates:
- TCC stubs (linux_ipc_stubs, ipc_overrides, claude-native) return denied
- getBridgeConsent returns consented: false
- getMediaAccessStatus returns 'denied'
- getOpenWindows returns empty array (no desktop enumeration)

Allowlist-only filesystem access:
- readLocalFile/openLocalFile/showInFolder restricted to home dir + /tmp
- No deny-lists — uses isPathWithinAllowedRoots with realpathSync

Input validation and resource limits:
- --sdk-url restricted to api.anthropic.com over HTTPS only
- claude:// protocol URLs validated before forwarding
- $TERMINAL validated against system bin directories only
- stdinHistory capped at 10MB with oldest-first eviction
- Trace log rotation at 50MB

Infrastructure hardening:
- --inspect bound to 127.0.0.1 (not 0.0.0.0)
- --no-sandbox conditional on user namespace availability
- Transcript recovery prompts sanitized against stored prompt injection

Tests updated to validate secure behavior (387 pass, 0 fail).

https://claude.ai/code/session_01E4BVBGwGTnwPNXXBL2bLnd

PR #89, opened 2026-04-23
security: extract focused hardening hunks from #75 for v4.3.2

johnzfitch:
Extracts the security-hardening portion of #75 (the `8969a90` commit) without the accompanying refactor, so the defensive posture can ship as a targeted v4.3.2 patch instead of waiting on the full session/transcript/resume rewrite.

#75 stays open for the v4.4.0 release (refactor + permission-manager infrastructure).

## Scope — what's in

**Deny-by-default permission stubs**
- `@ant/claude-native` — `isAccessibilityEnabled` / `hasScreenCapturePermission` / both request variants now return `false` / `Promise.resolve(false)` (Linux has no TCC prompt; auto-granting was a lie).
- `frame-fix-wrapper` — `systemPreferences.getMediaAccessStatus` returns `'denied'`; `askForMediaAccess` returns `false`.
- `ipc_overrides` — `getBridgeConsent` now returns `{consented: false}` (aligns with keeping dispatch disabled on the Linux port — see rationale in PR body).

**Allowlist-only filesystem access**
- `ipc_overrides` — `readLocalFile` / `openLocalFile` / `showInFolder` gated by new `isPathWithinAllowedRoots()` using `fs.realpathSync` + `path.sep` prefix check against `[homedir, /tmp]`. No deny-list; the allowlist IS the policy. Symlink-resistant.

**Input validation**
- `session_orchestrator` — `--sdk-url` restricted to `api.anthropic.com` (or subdomain) over HTTPS. Malformed / non-matching URLs dropped from the bridge spawn args instead of passed through.
- `frame-fix-wrapper` — `claude://` URLs parsed as `URL`, protocol checked, length-capped at 2048, rejected if they contain `< > " ' \``. Malformed URLs rejected with a warning, not forwarded.
- `ipc_overrides` — `$TERMINAL` resolved via `execFile('which', …)` and validated against `/usr/bin/`, `/usr/local/bin/`, `/usr/lib/`, `/snap/bin/` prefixes. Shell-profile poisoning no longer turns into code execution. `execFile` options tightened with a 3s timeout and stderr `ignore`.

**Resource / retention limits**
- `claude-swift` — `stdinHistory` capped at 10 MB with oldest-first eviction down to 50% watermark.
- `claude-swift` — trace log file rotated when it exceeds 50 MB; keeps last 10 MB.
- `claude-swift` — `getOpenWindows()` returns `[]` (no `wmctrl` enumeration of the user's desktop).

**Infrastructure hardening**
- `launch.sh` — `--no-sandbox` is now conditional on user-namespace availability (`/proc/sys/kernel/unprivileged_userns_clone` or `/proc/sys/user/max_user_namespaces`). Chromium renderer sandbox is re-enabled on distros that support it.
- `launch.sh` — `--inspect=9229` → `--inspect=127.0.0.1:9229` when `--perf` is set (defense-in-depth; Node already defaults to localhost, but explicit is better).

**Transcript recovery prompt injection defense**
- `transcript_store` — new `sanitizeTranscriptForRecovery()` strips the recovery prompt's structural markers (`[Local cowork continuity recovery]`, `New user message:`, `Recent conversation:`) from transcript text before re-injecting into a recovery prompt. Prevents stored prompt injection where prior AI output could manipulate recovery framing.

## Scope — what's deliberately out (staying in #75 → v4.4.0)

- `stubs/cowork/permission_manager.js` (new 212-line user-prompt infrastructure) — the deny-by-default changes in this PR do not depend on it. If / when the prompt flow lands in v4.4.0, the stubs here get replaced with `permissionManager.requestWithDialog(...)` calls.
- Removal of `resume_coordinator.js`, rewrite of `session_store.js` (-692), rewrite of `transcript_store.js` (-415 beyond the sanitize hook) — all cleanup/refactor, not security.
- Enabling dispatch (`ipc_overrides` bridge handler removal, swift stub `setGuestRequestCallback` / `sendGuestResponse` from #72) — project stance is to keep dispatch disabled on the Linux port.

## Tests

- Baseline (post-#87 master): 358 pass, 0 fail.
- After this PR: **387 pass, 0 fail** (+29 from new `tests/node/current-path/security_hardening.test.cjs`).
- `node -c` syntax check on all modified JS files: clean.

## Limits of testing in this prep env

Headless — no Electron GUI run, no Wayland compositor, no `./install.sh --doctor`. Recommend a manual smoke after merge:
- Launch on a userns-capable distro and confirm `"User namespaces available — Chromium sandbox enabled"` appears; confirm UI still works under sandbox.
- Verify a `readLocalFile` IPC call for `/etc/passwd` gets the `BLOCKED (outside allowed roots)` warn and returns `null`.
- Trigger `claude://<valid>` and confirm it's forwarded; confirm a URL with `<` in it is rejected.

## Commit

Single cherry-pick of `8969a90` from #75 (`Fix 13 security vulnerabilities left by misaligned AI contributor`). No conflicts; applied cleanly on top of `cbb5b1d` (post-v4.3.1 master).

---
_Generated by [Claude Code](https://claude.ai/code/session_014fH9BxzhXW4KVvW3efguHL)_

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -231,10 +231,27 @@ if ! dbus-send --session --print-reply --dest=org.freedesktop.DBus /org/freedesk
   PASSWORD_STORE="basic"
 fi
 
+# Sandbox check: prefer Chromium sandbox when unprivileged user namespaces
+# are available. --no-sandbox is required on distros without userns support
+# (disables Chromium renderer process isolation — see security notes).
+_sandbox_flag="--no-sandbox"
+if [[ -f /proc/sys/kernel/unprivileged_userns_clone ]]; then
+  if [[ "$(cat /proc/sys/kernel/unprivileged_userns_clone 2>/dev/null)" == "1" ]]; then
+    _sandbox_flag=""
+    echo "User namespaces available — Chromium sandbox enabled"
+  fi
+elif [[ -f /proc/sys/user/max_user_namespaces ]]; then
+  _max_ns="$(cat /proc/sys/user/max_user_namespaces 2>/dev/null)"
+  if [[ "$_max_ns" -gt 0 ]] 2>/dev/null; then
+    _sandbox_flag=""
+    echo "User namespaces available — Chromium sandbox enabled"
+  fi
+fi
+
 # Build electron args
 _electron_args=(
   "./${ASAR_FILE}"
-  --no-sandbox
+  ${_sandbox_flag}
   --password-store="$PASSWORD_STORE"
   --enable-features=GlobalShortcutsPortal
   --class=Claude
@@ -245,8 +262,9 @@ if [[ "$_perf" == true ]]; then
   export CLAUDE_COWORK_IPC_TAP=1
   export CLAUDE_COWORK_TRACE_IO=1
   export CLAUDE_COWORK_VERBOSE=1
+  # SECURITY: Bind to localhost only — never expose to network
   _electron_args+=(
-    --inspect=9229
+    --inspect=127.0.0.1:9229
     --remote-debugging-port=9222
   )
   echo ""
--- a/stubs/@ant/claude-native/index.js
+++ b/stubs/@ant/claude-native/index.js
@@ -195,13 +195,13 @@ const nativeStub = {
     spawn('xdg-open', [revealDir], { detached: true, stdio: 'ignore' });
   },
 
-  // Accessibility
-  isAccessibilityEnabled: () => true,
-  requestAccessibilityPermission: () => Promise.resolve(true),
+  // Accessibility — deny by default (Linux has no TCC prompt)
+  isAccessibilityEnabled: () => false,
+  requestAccessibilityPermission: () => Promise.resolve(false),
 
-  // Screen capture
-  hasScreenCapturePermission: () => true,
-  requestScreenCapturePermission: () => Promise.resolve(true),
+  // Screen capture — deny by default (Linux has no TCC prompt)
+  hasScreenCapturePermission: () => false,
+  requestScreenCapturePermission: () => Promise.resolve(false),
 };
 
 // ============================================================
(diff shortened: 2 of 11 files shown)


johnzfitch:
Addressed all 6 review comments in `e864548`. Summary:

| # | Reviewer | Comment | Fix |
|---|----------|---------|-----|
| 1 | CodeQL | Incomplete URL substring sanitization in `security_hardening.test.cjs:162` | Replaced `args.includes('https://api.anthropic.com')` with strict array-position check (`args[indexOf('--sdk-url') + 1] === inputUrl`). Stronger assertion + dodges the false positive. |
| 2 | Copilot | Log injection / credential leak via `sdkUrl` in malformed-URL warn (`session_orchestrator.js:343`) | Dropped the URL from the log line; generic message only. |
| 3 | Copilot | `decodeURIComponent` throw → renderer-triggered DoS (`ipc_overrides.js:252`) | Wrapped in `try/catch` in all three handlers (`readLocalFile`, `openLocalFile`, `showInFolder`). Decode failures return the same null/void as a denied path. |
| 4 | Copilot | Allowlist comment mentions "sessions base" but code only includes `[homedir, /tmp]` (`ipc_overrides.js:34`) | Updated the comment to match the policy. Did not extend the allowlist to `DIRS.claudeLocalAgentRoot` because the default XDG paths already live under `homedir`; if anyone sets `XDG_CONFIG_HOME` outside home, that's a separate decision worth its own PR. |
| 5 | Copilot | `showInFolder` missing `path.isAbsolute` check that the other two handlers have (`ipc_overrides.js:295`) | Added `path.isAbsolute` gate to `showInFolder`, **and** hardened `isPathWithinAllowedRoots` itself to reject non-absolute inputs (so policy is never `process.cwd`-dependent via `path.resolve` fallback). The latter is the belt-and-suspenders fix. |
| 6 | Copilot | stdin cap bypass when single chunk > MAX (`claude-swift/js/index.js:1864`) | Special-case oversized chunks: clear history, skip recording. The previous eviction loop emptied the buffer but still pushed the oversized chunk, leaving `stdinHistoryBytes` permanently above the cap. |

Tests: **387 pass, 0 fail** (no test count change — fixes are behavioral, the existing tests already cover them).

Not resolving the threads programmatically — leaving that for you to one-click resolve as you re-review.

---

PR #90, opened 2026-04-23
Fixes Issue #88 (cowork) - Translate VM paths in spawn() args and env-var values

### LINKED ISSUE #88: Cowork Path Translation Issue (Works in end, but wasteful repeated attempts)? NEEDS INPUT
### Area

Path translation

### What happened?

Issue: cowork path translation doesn't cover bash command arguments when running via PKGBUILD/aur

## Symptom

When Cowork sessions on Linux spawn `bash -c "..."` calls that reference the
macOS-style VM path layout (`/sessions/<sessionName>/mnt/<key>/...`), the paths
reach the shell verbatim and fail with `No such file or directory`. Claude
typically retries 1–3 times with different path guesses before stumbling onto
the real host path or asking the user. 

## Repro

Install via PKGBUILD/AUR
Drop a file into a directory you've shared with Cowork (e.g., `~/Documents/Claude/foo.xlsx`)
and ask Claude to read it. Watch
`~/.local/state/claude-cowork/logs/claude-swift-trace.log` — saw entries like:

```
[TRACE] vm.spawn() command=bash
[TRACE] vm.spawn() args=["-c","ls -la /sessions/trusting-brave-newton/mnt/Claude"]
[TRACE] vm.spawn() additionalMounts={"Claude":{"path":"home/xyz/Documents/Claude","mode":"rw"}, ...}
```

The arg `/sessions/trusting-brave-newton/mnt/Claude` is what the bundled SDK
constructs from its VM mental model; on Linux without a VM, that path doesn't exist.

## Why it happens

`stubs/cowork/asar_adapter.js` already ships path-rewriting infrastructure
(`DEFAULT_FILESYSTEM_PATH_ALIASES`, `isFileSystemPathRewriteChannel`,
`rewriteAliasedFilePath`, etc.) and it works for IPC channels — file open
dialogs, preview generation, and similar operations transparently translate
VM paths to host paths.

That same translation does not appear to be applied to `vm.spawn` command
arguments. So shell tool calls bypass the rewriter entirely and the SDK's
VM-style paths reach `/usr/bin/bash` unchanged.

## Current Known Caveat
Mentioned/Addressed in PR attached

## Suggested fix

Extend `session_orchestrator.js`'s `vm.spawn` (or the equivalent translation
layer it calls) to rewrite occurrences of `/sessions/<sessionName>/mnt/<key>/...`
in each `args[]` entry, using the per-session `additionalMounts` dict already
in scope. 

## Severity

**Low.** Claude self-corrects on failed paths — it's an annoyance and a bit of
wasted context, not a blocker. Worth fixing for polish but not urgent.

### Steps to reproduce

1. Install Claude Cowork on Linux via AUR (or any setup where
   frame-fix-entry.js is wired up as the asar's main aside from sudo symlink caveat).
2. Add a directory to Cowork's allowed mounts (e.g., ~/Documents/Claude)
   via the app's file/folder permission UI on first session start.
3. Drop a file into that shared directory (e.g., ~/Documents/Claude/test.xlsx).
4. In a Cowork session, ask Claude to read or list that file
   (e.g., "what's in test.xlsx").
5. Watch ~/.local/state/claude-cowork/logs/claude-swift-trace.log.

Expected: Claude finds and reads the file without failed attempts

Actual: Trace shows 3+ failed vm.spawn calls with macOS-VM-style paths
like `bash -c "ls /sessions/<sessionName>/mnt/Claude/..."` returning ENOENT,
followed by Claude retrying with different path guesses until it stumbles
onto the correct host path or asks the user.

Example trace excerpt:
  [TRACE] vm.spawn() command=bash
  [TRACE] vm.spawn() args=["-c","ls -la /sessions/trusting-brave-newton/mnt/Claude"]
  [TRACE] vm.spawn() additionalMounts={"Claude":{"path":"home/<user>/Documents/Claude","mode":"rw"}, ...}
  -> ls: cannot access '/sessions/trusting-brave-newton/mnt/Claude': No such file or directory

Arch Linux / KDE Plasma 6 / Wayland

fcd8111dcb65269cdd82e1347d1177d07a6ea3f8

1.3883.0

### Checklist

- [x] I ran `./install.sh --doctor` and included its output if relevant (ran claude-cowork --doctor as from the AUR)
- [x] I checked the log file and redacted any tokens before pasting
- [x] I searched existing issues and this isn't a duplicate

F1nny:
Addresses and Closes #88

## What does this change?
The bundled SDK constructs `/sessions/<n>/mnt/<key>/...` paths from its macOS VM mental model and passes them to `vm.spawn()` as bash command args. On Linux there's no VM; those paths reach `bash` verbatim and ENOENT. Claude self-corrects by retrying 1–3 times with different path guesses — annoying for users, slow, and wasteful of context.

### Why
`canonicalizePathForHostAccess` in `stubs/cowork/dirs.js` already translates VM paths to host paths, and is wired up to all the IPC channels (filesystem.read/write/exists/stat/watch, etc.) plus spawn `cwd`. Spawn `args` and env-var values just weren't routed through it — almost certainly an oversight when the IPC translation was added. 

### Fix
Adds `translateVmPathsInString(sessionsBase, str)` — a regex-based helper that finds VM paths embedded in strings and rewrites them via the existing `canonicalizePathForHostAccess`. Applied to `spawn()` args + envVars values and `spawnSync()` args.
The regex stops at whitespace and shell metacharacters, so paths in `bash -c "..."` strings (quoted, piped, redirected, subshelled) translate cleanly. Mount keys with shared prefixes (e.g. Claude vs Claude-archive) don't collide.

## Type

- [X] Bug fix
- [X] Distro / DE compatibility
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing
9/9 unit cases pass against the real exported helper:
- Translates: bare paths, quoted paths, piped/redirected paths, multiple paths in one arg, command-substitution paths
- No-ops on: host paths, non-path args, paths containing /.. (3 traversal cases)

Manual repro:
1. Build with this branch + PR #87's PKGBUILD wiring (and/or with latest commit cbb5b1d3ce4e6324620de7f4084a3624f6ae49c5)
2. Launch Cowork, share ~/Documents/Claude, ask Claude to read a file
3. Check ~/.local/state/claude-cowork/logs/claude-swift-trace.log — should show vm.spawn() args=[..., 'ls /home/.../Claude/...'] on the first attempt rather than /sessions/... failures.

- Distro / desktop: Arch Linux / KDE Plasma 6 Wayland
- Tested with `./install.sh --doctor`: X
- Tested a full Cowork session: X

## Security impact

**- [x] This change does touch security-sensitive code — explanation below:**

This PR modifies `spawn()` and `spawnSync()` in `stubs/@ant/claude-swift/js/index.js` to translate `/sessions/<n>/mnt/<key>/...` VM paths in command args (and env-var values for `spawn()`) using the existing `canonicalizePathForHostAccess` infrastructure already used for IPC channels and spawn `cwd`.

**Alignment with OAUTH-COMPLIANCE.md:**

- **`filterEnv` not modified.** Filtered env vars enter `spawn()` after `filterEnv` runs; this PR rewrites `/sessions/...` substrings in their values but never adds, removes, or reorders vars. OAuth/API tokens are opaque and don't match the VM-path pattern, so they pass through untouched.
- **`AuthRequest` not modified.** Auth flow (`stubs/@ant/claude-native/index.js`, `ALLOWED_AUTH_ORIGINS`, browser-only handoff) is not in this PR's scope.
- **`isPathSafe` not modified.** The new helper transitively uses it through the existing `canonicalizePathForHostAccess` → `translateVmPathStrict` chain — same chain every current IPC translation callsite uses.

**Boundary preservation (the security delta this PR introduces):**

Without translation, `bash -c "ls /sessions/X/mnt/Claude/../etc"` ENOENTs because the literal `/sessions/` path doesn't exist on Linux — the user-granted mount boundary is preserved by accident. A naive translation would resolve through the host symlink and escape the boundary. This PR **refuses to translate any matched VM path containing `/..`** (`if (match.includes('/..')) return match;`), so traversal-style paths stay untranslated and bash ENOENTs them just like today, restoring the macOS VM guarantee.

**Pre-existing weakness flagged but out of scope:**

`translateVmPathStrict` permits `/sessions/X/mnt/Y/../../../etc/passwd` — `path.normalize` collapses it cleanly to `etc/passwd`, bypassing the `startsWith('..' + sep)` check, resolving to `<sessionsBase>/etc/passwd`. Inherited by every existing caller (IPC channels included); not introduced by this PR. Worth a separate hardening PR. The `/..` refusal here defends spawn args from it directly.

## Compatibility
Pure addition. Host paths pass through. Non-string args pass through. Failed translations fall back to the arg as-is. No changes to existing functions. macOS unaffected.

## Relationship to the documented /sessions root symlink approach:
The README's "Known caveats" section documents install.sh's solution — sudo ln -s "$HOME/.config/Claude/local-agent-mode-sessions/sessions" /sessions — which makes the OS resolve VM paths natively. That works great for install.sh users but doesn't work for the AUR users. This PR provides the same end result via stub-level translation: same final paths, no /sessions filesystem entry required. Compatible with users who also have the root symlink — the explicit translation just produces the same path the OS would resolve to, no conflict.

## Checklist

- [X] No API keys, tokens, `.env` files, or log output with credentials included
- [X] Commit messages follow the project style (no emoji, brief summary + explanation)
- [X] `./install.sh --doctor` passes cleanly on my system

### DIFF
--- a/stubs/@ant/claude-swift/js/index.js
+++ b/stubs/@ant/claude-swift/js/index.js
@@ -40,6 +40,7 @@ const {
   canonicalizeHostPath,
   canonicalizeVmPathStrict: _canonicalizeVmPathStrict,
   canonicalizePathForHostAccess: _canonicalizePathForHostAccess,
+  translateVmPathsInString: _translateVmPathsInString,
 } = require('../../../../cowork/dirs.js');
 const { createSessionStore } = require('../../../../cowork/session_store.js');
 const { createSessionsApi } = require('../../../../cowork/sessions_api.js');
@@ -146,6 +147,15 @@ function translateVmPathStrict(vmPath) {
     }
     throw err;
   }
+}
+
+// Wrapper over dirs.js translateVmPathsInString that binds SESSIONS_BASE.
+// Used to rewrite VM paths embedded in spawn() command args and env-var
+// values (the bundled SDK constructs /sessions/<n>/mnt/<key>/... paths
+// from its macOS VM mental model; on Linux those paths reach bash verbatim
+// and ENOENT without translation).
+function translateVmPathsInString(str) {
+  return _translateVmPathsInString(SESSIONS_BASE, str);
 }
 
 function extractSessionNameFromVmPathStrict(vmPath) {
@@ -1487,13 +1497,26 @@ class SwiftAddonStub extends EventEmitter {
       trace('spawn envVars keys from asar: ' + Object.keys(envVars).join(', '));
     }
     try {
+      // Translate /sessions/<n>/mnt/<key>/... VM paths embedded in args
+      // and env-var values to host paths. The bundled SDK constructs these
+      // from its macOS VM mental model; on Linux they reach bash verbatim
+      // and ENOENT without translation. cwd is already translated downstream
+      // in buildSpawnOptions; this covers the remaining surfaces.
+      const translatedArgs = Array.isArray(args)
+        ? args.map(translateVmPathsInString)
+        : [];
+      const translatedEnvVars = envVars && typeof envVars === 'object'
+        ? Object.fromEntries(
+            Object.entries(envVars).map(([k, v]) => [k, translateVmPathsInString(v)])
+          )
+        : {};
       const processState = {
         id,
         processName,
         command,
-        args: Array.isArray(args) ? args.slice() : [],
+        args: translatedArgs,
         options: options && typeof options === 'object' ? { ...options } : {},
-        envVars: envVars && typeof envVars === 'object' ? { ...envVars } : {},
+        envVars: translatedEnvVars,
         additionalMounts,
         allowedDomains,
         sharedCwdPath,
@@ -1528,7 +1551,9 @@ class SwiftAddonStub extends EventEmitter {
       if (optEnv && typeof optEnv === 'object') {
         trace('WARNING: spawnSync() ignoring options.env override');
       }
-      const result = nodeSpawnSync(command, args || [], { encoding: 'utf-8', cwd, ...safeOptions });
+      // Translate VM paths in args (mirrors the same fix in spawn() above).
+      const translatedArgs = Array.isArray(args) ? args.map(translateVmPathsInString) : [];
+      const result = nodeSpawnSync(command, translatedArgs, { encoding: 'utf-8', cwd, ...safeOptions });
       return { stdout: result.stdout, stderr: result.stderr, status: result.status, signal: result.signal, error: result.error };
     } catch (err) {
       console.error('[claude-swift] spawnSync error:', err);
--- a/stubs/@ant/claude-swift/js/index.js
+++ b/stubs/@ant/claude-swift/js/index.js
@@ -548,6 +548,25 @@ function extractSessionName(args, envVars, sharedCwdPath) {
         return sessionName;
       }
     }
+
+    // Fallback: detect /sessions/<name>/mnt/ patterns embedded inside arg
+    // strings. The SDK's bash -c "..." invocations pack VM paths into the
+    // shell command body rather than as standalone args, so the whole-arg
+    // checks above miss them. We round-trip through the strict validator
+    // to defend against `..`-style session names hijacking createMountSymlinks
+    // (e.g. /sessions/../../../etc/passwd/mnt/...).
+    for (const arg of args) {
+      if (typeof arg !== 'string') continue;
+      const m = arg.match(/\/sessions\/[^\/\s'"`<>|&;()$\n]+\/mnt(?=\/|$|\s|['"`<>|&;()$])/);
+      if (!m) continue;
+      try {
+        const sessionName = extractSessionNameFromVmPathStrict(m[0]);
+        trace('Extracted session name from embedded VM path in arg: ' + sessionName);
+        return sessionName;
+      } catch (err) {
+        trace('SECURITY: Rejected embedded VM path while extracting session name: ' + err.message);
+      }
+    }
   }
 
   trace('WARNING: No validated VM path available for session name extraction');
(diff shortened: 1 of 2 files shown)


F1nny:
Standby found edge working on

F1nny:
Reopening with a follow-up commit (16396b1) added after testing revealed a second gap: even with VM paths translated in spawn args, findSessionName couldn't extract the session name from bash -c "..." strings, so createMountSymlinks never ran and the translated paths still ENOENTed. Both commits together complete the chain.

F1nny:
Heads-up unrelated to this PR: noticed the "Publish to AUR" step has been failing on every PKGBUILD-touching merge since #78 (also failed for #87 twice). Last successful AUR commit on the AUR repo seems to be e6a1140 (the merge of #61) 
May be an AUR_SSH_KEY issue but could only confirm with action logs or you may see it already - Happy to help diagnose if can :heart:

johnzfitch:
Closing in favor of #102 which incorporates your VM path translation approach.

Your fixes that landed:
- `translateVmPathsInString()` in `dirs.js` &mdash; regex-based rewriting of `/sessions/<name>/mnt/<key>/...` paths embedded in strings
- Applied to `spawn()` args, env var values, and `spawnSync()` args
- `/..` traversal refusal preserved from your original implementation

The regex, shell metacharacter boundary, and fallback-on-failure design are directly from your PR. #91 (nested mount mkdir) was merged separately as-is.

Thank you @F1nny for the thorough analysis in #88 and both PRs &mdash; the path translation fix eliminates the retry loops that were wasting context on every Cowork session.

---

F1nny:
Thanks @johnzfitch ! Looking great

Issue #97, opened 2026-04-26
Unsupported platform: linux-x64 thrown from getHostPlatform() during session init; cascading binary/version mismatches downstream


Crono141:
### Area

Cowork session (spawn / spawn fails)

## Description

On a fresh install against current Claude Desktop builds (SDK 2.1.119), prompt submission fails with `Unsupported platform: linux-x64`. The existing platform-gate patch in `enable-cowork.py` is being applied successfully, but the error originates from a different function (`getHostPlatform()`) that's called downstream during `prepare()` on every session init.

After patching that, two further issues surface in sequence — one related to platform spoofing causing a Mach-O binary to be downloaded, and one related to CLI version drift. Documenting all three together since they're effectively a single broken pipeline.

## Environment

- **Distro:** CachyOS (Arch-based)
- **claude-cowork-linux:** installed via [method — AUR/install.sh/etc.]
- **Claude Desktop version:** 1.1.4498
- **claude-code SDK version (downloaded):** 2.1.119
- **Local `claude` binary version (initial):** 2.1.91

## Repro

1. Install per README.
2. Launch `claude-cowork`, sign in, point at a working folder.
3. Submit any prompt.
4. Session fails to initialize.

## Issue 1: `getHostPlatform()` throws on Linux

`enable-cowork.py` patches the platform-availability gate and applies cleanly (`/*cowork-patched*/` marker present). However, `prepare()` calls `getHostTarget()` which calls `getHostPlatform()`, and that function has its own platform check that only handles `darwin` and `win32`:

```js
getHostPlatform(){const A=process.arch;if(process.platform==="darwin")return A==="arm64"?"darwin-arm64":"darwin-x64";if(process.platform==="win32")return A==="arm64"?"win32-arm64":"win32-x64";throw new Error(`Unsupported platform: ${process.platform}-${A}`)}
```

Stack trace from `~/.config/Claude/logs/main.log`:

```
Error: Unsupported platform: linux-x64
    at HSr.getHostPlatform (/usr/lib/claude-cowork/app.asar/.vite/build/index.js:468:58573)
    at HSr.getHostTarget (...:468:58668)
    at HSr.prepare (...:468:65575)
    at async eP.doSessionInitialization (...:5951:62561)
```

[...]

PR #98, opened 2026-05-01
fix: Session init issue #97 

michael-greider:
Fixes issue #97 all three items, additionally fixed a crash that surfaced from fixing item 1

Commit 1:
-Patched getHostPlatform() throw
-Patched Eo.files[platform][arch] 
-installSdk(): ELF binary from CDN works over no-op 
-launch.sh: CDN download instead of sym  link for claude-code/ and claude-code-vm/ *doesn't touch your claude code install just the expected one from cowork*
-launch.sh: clear V8 byte cache before launch *this does not touch session files only electron cache that was breaking people's installation on updates*

Commit 2: Split from commit 1 - i decided I was a POS for making the initial PR AUR only 

These resolve:

1. getHostPlatform() throwing linux-x64
2. Mach-O binary downloaded | launcher replaces remaining Mach-O 
3. --effort xhigh rejected | only needed the binary in claude-code-vm
4. Eo.files[linux] undefined 

Testing:

-Clean slate rm -rf ~/.config/Claude/claude-code{,-vm} | I have the same dev environment as the repo owner but I have confirmed the patches via running the install on my main driver and my sandbox pc that had no cowork install

-Ran cowork for a day after applying these fixes, applied the non AUR only ones, uninstalled, then reinstalled- my sessions stayed 

-no TypeErrors in logs, ran from terminal to monitor 

- '--doctor' reported no Mach-O warnings  

Commit 3: 

- upstream did not suppress EBADF, presented stability issues

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -9,6 +9,9 @@ license=('custom:proprietary')
 depends=(
     'electron'
     'nodejs'
+    'curl'
+    'zstd'
+    'file'
 )
 # gnome-keyring is recommended but not required; launcher detects SecretService
 # at runtime and falls back to --password-store=basic if unavailable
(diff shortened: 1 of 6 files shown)


michael-greider (commented):
Set ad DRAFT ** punycode node update has changed the fix. approach.

michael-greider:
Alright I hit full coverage on issue #97 and extended the fix from AUR only as to not be a POS

johnzfitch:
Closing in favor of #102 which incorporates your `getHostPlatform` patch and `installSdk` implementation with additional validation.

Your fixes that landed:
- `enable-cowork.py`: `getHostPlatform()` throw replaced with `return"darwin-x64"`
- `installSdk()`: real implementation that downloads Linux binary from CDN

Additional hardening in #102:
- `installSdk` validates `subpath` resolves within `~/.config/Claude/` before any filesystem operations
- Uses `execFileSync('curl', [...])` and `execFileSync('zstd', [...])` separately instead of `sh -c` piping
- `version` is URI-encoded in the download URL

Thank you @michael-greider for the thorough three-part diagnosis in #97 and the working fix &mdash; this unblocks Linux users on current Claude Desktop builds.

---

PR #100, opened 2026-05-05
fix: OAuth claude:// callback now completes login on Linux

mrbobbytables:
## What does this change?

**NOTE:** This was fully generated by Claude. It did not work out of the box on  [Bluefin](https://projectbluefin.io/) as the root of the drive is RO and it could not create the `/sessions` to mirror the OSX structure.

After getting it up and running, it'd launch - but there were a lot of issues with the auth-flow callback not working correctly.  After awhile of iterating and testing it now works without issue.

The below  description is generated from Claude:

Three bugs prevented the auth flow from working:

1. The claude-desktop protocol handler was running launch.sh for claude:// callbacks, which repacks the asar using 'npx asar' (needs node in PATH). xdg-open invokes handlers with a restricted PATH where node isn't present, so the repack failed with 'env: node: No such file or directory' and the Electron process died before requestSingleInstanceLock() could run.

2. frame-fix-wrapper.js called requestSingleInstanceLock() before index.js, but with app name '@ant/desktop' (from package.json) instead of 'Claude'. This put the lock at ~/.config/@ant/desktop while index.js put its own lock at ~/.config/Claude — two different locks, so second-instance never fired. index.js's lock call returned false and it called app.quit(), killing the app.

3. The electron wrapper at ~/.local/bin/electron is a Node.js script that also requires node in PATH. The forwarder was invoking it via the wrapper.

Fixes:
- Add protocol-forwarder.js: minimal Electron entry that only calls requestSingleInstanceLock() to trigger second-instance on the running app, then quits. Falls back to launching the full app via launch.sh if no instance is running (handles the race where the user starts auth before launching app).
- Remove requestSingleInstanceLock() from frame-fix-wrapper.js entirely — the asar's own lock+handler in index.js already handles second-instance correctly, including forwarding claude:// URLs to the claudeURLHandler (bEe/fCA).
- Call app.setName('Claude') in frame-fix-wrapper.js before index.js runs, so index.js's requestSingleInstanceLock() uses ~/.config/Claude (correct path).
- Use the raw ELF electron binary (~/.local/lib/node_modules/electron/dist/electron) instead of the node wrapper script when spawning the forwarder from the claude-desktop protocol handler — no node in PATH required.

## Type

- [x] Bug fix
- [x] Distro / DE compatibility
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

- Distro / desktop: [Bluefin](https://projectbluefin.io/)
- Tested with `./install.sh --doctor`: yes
- Tested a full Cowork session: yes

## Security impact

Please see the above message

## Checklist

- [ x] `./install.sh --doctor` passes cleanly on my system

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -515,7 +515,7 @@ install_stubs() {
     cp "$native_src" "$target_dir/node_modules/@ant/claude-native/index.js"
 
     # Copy frame-fix files
-    for f in frame-fix-wrapper.js frame-fix-entry.js; do
+    for f in frame-fix-wrapper.js frame-fix-entry.js protocol-forwarder.js; do
         if [[ -f "$stub_src/stubs/frame-fix/$f" ]]; then
             cp "$stub_src/stubs/frame-fix/$f" "$target_dir/$f"
         fi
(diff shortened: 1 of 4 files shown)


johnzfitch:
Still needed for the OAuth leg of #104. The single-instance-lock app-name fix and the protocol-forwarder are both real. Needs a rebase onto current master (v5.1.0 changed frame-fix-wrapper.js significantly). Can you confirm this still works on Bluefin after rebasing?

---

mrbobbytables:
PR has been rebased and with some additional fixes works again on bluefin :+1:

johnzfitch:
Closing as superseded by #127, now merged to `master`. #127 explicitly consolidates this PR — the OAuth `claude://` callback fix landed there as `stubs/frame-fix/protocol-forwarder.js` (URL forwarding with validation), a simplified single-instance lock, and a `/sessions` fs path redirect so the root symlink is no longer required.

---

mrbobbytables:
Thanks!

Issue #66, opened 2026-04-01
Lauch failure with Uncaught Exception: TypeError: xe.systemPreferences.setUserDefault is not a function


bradydonald:
### Area

Install / setup

### What happened?

Was working previously but decided to update to v4.3.0 and now doesn't launch with error:

```
A JavaScript error occurred in the main process
Uncaught Exception:
TypeError: xe.systemPreferences.setUserDefault is not a function
    at Object.<anonymous> (/home/d/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:5046:105350)
    at Module._compile (node:internal/modules/cjs/loader:1820:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1953:10)
    at Module.load (node:internal/modules/cjs/loader:1540:32)
    at Module._load (node:internal/modules/cjs/loader:1342:12)
    at c._load (node:electron/js2c/node_init:2:17999)
    at wrapModuleLoad (node:internal/modules/cjs/loader:262:19)
    at Module.require (node:internal/modules/cjs/loader:1563:12)
    at Module.require (/home/d/.local/share/claude-desktop/.asar-cache/app.asar/frame-fix-wrapper.js:847:34)
    at require (node:internal/modules/helpers:152:16)
```

This is on Fedora 43, Wayland, KDE

```
fastfetch
             .',;::::;,'.                 d@thinkpad
         .';:cccccccccccc:;,.             ----------
      .;cccccccccccccccccccccc;.          OS: Fedora Linux 43 (KDE Plasma Desktop Edition) x86_64
    .:cccccccccccccccccccccccccc:.        Host: 21NS0013US (ThinkPad X1 Carbon Gen 13)
  .;ccccccccccccc;.:dddl:.;ccccccc;.      Kernel: Linux 6.19.10-200.fc43.x86_64
 .:ccccccccccccc;OWMKOOXMWd;ccccccc:.     Uptime: 1 day, 16 hours
.:ccccccccccccc;KMMc;cc;xMMc;ccccccc:.    Packages: 3643 (rpm), 59 (flatpak-system), 19 (flatpak-user)
,cccccccccccccc;MMM.;cc;;WW:;cccccccc,    Shell: zsh 5.9
:cccccccccccccc;MMM.;cccccccccccccccc:    Display (DELA0FA): 3440x1440 in 34", 100 Hz [External]
[...]

PR #103, opened 2026-05-06
fix(frame-fix): stub macOS-only app activity APIs on Linux

wlcarden:
## Summary

After updating to commit `1b8df05` (asar v1.569.0 compat), the app crashes
with `TypeError: CA.app.invalidateCurrentActivity is not a function` on
session navigation and any deeplink that resolves to a session route.

This is the same class of bug that #66 / `97fbe53` fixed for
`systemPreferences.setUserDefault` — a macOS-only Electron API called
from the asar under spoofed `darwin`. Applying the same stub pattern.

## Root cause

The bundled main process contains:

```js
function Vir(e){
  if(process.platform==="darwin"){
    if(!Jir()||!e){
      CA.app.invalidateCurrentActivity();
      return
    }
    CA.app.setUserActivity(Hir,{v:Kir,sessionId:e})
  }
}
```

`Vir()` is called from the URL/deeplink router (`wOn`) on every
`webContents` navigation. It's gated by `process.platform === "darwin"`
which would skip on a real Linux Electron — but `frame-fix-wrapper`
spoofs `process.platform` to `'darwin'` for app code, so this branch
runs on Linux and hits the missing `app.*` methods.

Both `app.invalidateCurrentActivity()` and `app.setUserActivity()` are
documented as macOS-only in the Electron API.

## Stack trace (for searchability)

\`\`\`
A JavaScript error occurred in the main process
Uncaught Exception:
  at t  (.../app.asar/.vite/build/index.js:7031:16971)
  at WebContents.<anonymous> (.../app.asar/.vite/build/index.js:7031:17067)
\`\`\`

## Fix

Stub both methods as no-ops at module load, immediately after the
\`systemPreferences\` block from #67. 16 lines, single file. Pattern
mirrors \`97fbe53\` exactly so future macOS-only API regressions follow
a consistent template.

## Test plan

- [x] \`node --test tests/node/current-path/*.test.cjs\` — 403 pass, 0 fail
- [x] \`node --check stubs/frame-fix/frame-fix-wrapper.js\` — syntax OK
- [x] Manually verified on Linux Mint / X11 (Cinnamon, PipeWire): clicking session entries that previously triggered the dialog now navigates cleanly; no regressions in normal navigation, login, or \`--doctor\` (17 passed, 1 unrelated PATH warn).

### DIFF
--- a/stubs/frame-fix/frame-fix-wrapper.js
+++ b/stubs/frame-fix/frame-fix-wrapper.js
@@ -31,6 +31,22 @@ if (typeof systemPreferences.getUserDefault !== 'function') {
   };
 }
 
+// Patch macOS-only app activity methods that don't exist on Linux.
+// frame-fix spoofs process.platform to 'darwin', so app code gated by
+// `process.platform === "darwin"` will call these on Linux. Without
+// stubs the main process crashes with "X is not a function".
+const { app } = require('electron');
+if (typeof app.invalidateCurrentActivity !== 'function') {
+  app.invalidateCurrentActivity = function() {
+    // no-op on Linux
+  };
+}
+if (typeof app.setUserActivity !== 'function') {
+  app.setUserActivity = function(type, userInfo, webpageURL) {
+    // no-op on Linux
+  };
+}
+
 // Inject frame fix and Cowork support before main app loads
 const Module = require('module');
 const originalRequire = Module.prototype.require;


johnzfitch:
Got rolled into #108 which merged today. Same activity stubs plus a few extras. Thanks for putting this together!

---

Issue #107, opened 2026-05-07
[BUG(S)] returning darwin-x64 uses 35 paths that are mac-os only out of 91 total (7 are non cross-compat)


michael-greider:
### Area

Path translation

### What happened?

#106 #104 Are the start to this and will pile up, the path translation as far is i have found is:
  - 49 ACTIVE — macOS code that executes because we spoof darwin                                       
  - 35 SKIPPED — Linux/fallback code that DOESN'T execute because we spoof darwin                      
  - 7 MACOS_API — direct calls to APIs that don't exist on Linux Electron    
  
To make it short my #98 fix was "f it i'll just make it work" and the latest commit is "let's make it work 100 percent properly", it is a catch-22 of either performance issues and tool uses lapsing (mine) or crashes that are unpredictable, in my case and the aforementioned issues it was on launch/login. But i highly suspect this will plague the project forever if it's not nipped in the bud with the #105 approach now it will always present itself.

### Steps to reproduce

Launching the application (as of right now) 

### Relevant log output

```shell
all 91 paths i could find
-01/91 RUNS | === darwin | pos:4520137
  A}function mLr(){const e=CA.app.getPath("crashDumps"),A=process.platform==="win32"?"reports":"completed",t=[We.join(e,A)];return process.platform==="darwin"&&t.push(We.join(e,"pending")),fLr(async()=>(await Awt(e).catch(i=>rt.error(i)),(await pLr(t)).filter(i=>i.endsWith(".dmp"))))}const ewt=(e={})=>{let A=e.maxMinidumpsPerSession||10,t,r,i;function n(){rt.log("Starting Electron crashReporter"),CA.crashReporter.start({companyName:"",ignoreSystemCrashHandler:!0,productName:CA.app.name||CA.app.ge


-02/91 RUNS | === darwin | pos:4529801
 PHardwareDataType[0].machine_model});return}}catch{}e({})})}catch{e({})}})}function ULr(){return process.platform==="win32"?LLr():process.platform==="darwin"?GLr():Promise.resolve({})}const owt=(e={})=>{const A={};let t=e.deviceModelManufacturer;const r={...bLr,...e};function i(){const s=CA.screen.getPrimaryDisplay(),o=Math.floor(s.size.width*s.scaleFactor),a=Math.floor(s.size.height*s.scaleFactor);A.screen_density=s.scaleFactor,A.screen_resolution=`${o}x${a}`}async function n(){const{manufactur


[...]

Issue #104, opened 2026-05-07
Cannot login to the app


Ashwin-droid:
### Area

Install / setup

### What happened?

<img width="830" height="761" alt="Image" src="https://github.com/user-attachments/assets/d11154f7-7287-450a-adf1-b8aff8d1fb02" />

<img width="830" height="761" alt="Image" src="https://github.com/user-attachments/assets/19909b83-2731-4565-be32-63b5ac518371" />

pretty much but then it launches fine

after launching nothing happens

<img width="830" height="761" alt="Image" src="https://github.com/user-attachments/assets/332be705-d989-4d57-ace1-cf6295ed910c" />

<img width="797" height="823" alt="Image" src="https://github.com/user-attachments/assets/48c18ef9-288b-4b90-b200-32395f186261" />
this does NOT open the app and neither is the login completed

### Steps to reproduce

1. Run launch.sh / run from desktop
2. click on "ok" twice on the errors shown above
3. click on get started
4. click on login with google

### Relevant log output

```shell
cat ~/.local/state/claude-cowork/logs/claude-swift-trace.log
[2026-05-07T06:32:33.227Z] === MODULE LOADING ===
[2026-05-07T06:32:33.227Z] Trace IO logging: disabled
[2026-05-07T06:32:33.228Z] Constructor START
[2026-05-07T06:32:33.228Z] Constructor COMPLETE. vm.setEventCallbacks=function
[2026-05-07T06:32:33.228Z] Instance created. vm=object vm.setEventCallbacks=function
[2026-05-07T06:32:33.228Z] Module exports set. default.vm.setEventCallbacks=function
[2026-05-07T06:32:33.273Z] quickAccess.dictation.setLanguage(en-US)
[2026-05-07T06:32:38.955Z] === MODULE LOADING ===
[2026-05-07T06:32:38.956Z] Trace IO logging: disabled
[2026-05-07T06:32:38.956Z] Constructor START
[2026-05-07T06:32:38.956Z] Constructor COMPLETE. vm.setEventCallbacks=function
[2026-05-07T06:32:38.957Z] Instance created. vm=object vm.setEventCallbacks=function
[2026-05-07T06:32:38.957Z] Module exports set. default.vm.setEventCallbacks=function
[2026-05-07T06:32:38.985Z] quickAccess.dictation.setLanguage(en-US)
[2026-05-07T06:32:40.416Z] quickAccess.overlay.setLoggedIn(true)
[...]

PR #108, opened 2026-05-07
feat: rails + recovery -- restore manual permission flow, explicit update flow, COMPAT.md matrix

### LINKED ISSUE #106: [BUG] TypeError: CA.app.invalidateCurrentActivity is not a function on startup
### Area

Launch (app won't start)

## Description
After updating Claude Desktop via `update-claude-cowork`, the app shows two JavaScript error dialogs on startup before loading successfully:
at WebContents.t (.../app.asar/.vite/build/index.js:7031:16971)

The error fires twice (once per WebContents window). After dismissing both dialogs, the app works normally.

## Root cause
The minified bundle calls `app.invalidateCurrentActivity()` and related macOS NSUserActivity/Handoff methods that don't exist in Electron on Linux.

## Environment
- **OS:** Zorin OS 18.1 (Ubuntu-based)
- **Arch:** x86_64
- **Electron:** system electron
- **Install method:** `./install.sh` from repo

## Workaround

Image
Image

Add a shim block in `launch.sh` that prepends no-op stubs to `index.js` before 
repacking. Insert this after the existing `resourcesPath` patch and before the 
`_needs_repack` check:

```bash
# Fix macOS-specific Electron app methods not available on Linux.
if [ -f "$INDEX_JS" ] && grep -q 'invalidateCurrentActivity' "$INDEX_JS" && ! grep -q '__LINUX_MACOS_SHIMS__' "$INDEX_JS"; then
  echo "Patching macOS app methods (invalidateCurrentActivity, etc.) for Linux..."
  _SHIM_CODE='/*__LINUX_MACOS_SHIMS__*/;(function(){var _a=require("electron").app;["invalidateCurrentActivity","setUserActivity","getCurrentActivityType","resignCurrentActivity","updateCurrentActivity"].forEach(function(m){if(typeof _a[m]!=="function")_a[m]=function(){};});})();'
  sed -i "1s|^|${_SHIM_CODE}|" "$INDEX_JS"
fi
```

The `__LINUX_MACOS_SHIMS__` marker prevents double-patching on subsequent launches.

### Steps to reproduce

1. Run `update-claude-cowork` to update to the latest Claude Desktop build
2. Launch with `claude-desktop`
3. Wait for the app to load (~30 seconds)
4. First error dialog appears: `TypeError: CA.app.invalidateCurrentActivity is not a function` (at `WebContents.t`)
5. Click "Ok"
6. Second error dialog appears: same TypeError (at `WebContents.<anonymous>`)
7. Click "Ok"
8. App loads normally after dismissing both dialogs

### Relevant log output

```shell
[MAIN_LOG] 2026-05-07 10:41:38 [warn] [CCD] Shell environment extraction failed, using process.env: Error: Shell path worker not found at: /home/cst/.config/Claude/cowork-resources/app.asar/.vite/build/shell-path-worker/shellPathWorker.js

[CONSOLE:16] "Error: Incoming "getInitialLocale" call on interface "DesktopIntl" from 'file:///...app.asar/.vite/renderer/main_window/index.html' did not pass origin validation"

[CONSOLE:16] "Error: Incoming "getInitialLocale" call on interface "DesktopIntl" from 'file:///...app.asar/.vite/renderer/find_in_page/find-in-page.html' did not pass origin validation"

[CONSOLE:115] "Uncaught ReferenceError: process is not defined", source: file:///...app.asar/.vite/renderer/main_window/assets/main-BG3-Qyem.js (115)

[CONSOLE:115] "Uncaught ReferenceError: process is not defined", source: file:///...app.asar/.vite/renderer/find_in_page/assets/main-DPJtB7nD.js (115)

[ERROR] GetVSyncParametersIfAvailable() failed (3 times)

[CONSOLE:0] "Error with Permissions-Policy header: Unrecognized feature: 'bluetooth'."
[CONSOLE:0] "Error with Permissions-Policy header: Unrecognized feature: 'web-share'."
```

Zorin OS 18.1 (Ubuntu 24.04 based), GNOME desktop (X11)

e04862b975ac3f33780c15225ff0e104f43a2e17 (2026-05-06)

1.6259.1

### Checklist

- [ ] I checked the log file and redacted any tokens before pasting
- [ ] I searched existing issues and this isn't a duplicate

johnzfitch:
## Summary

Two coherent tracks landing in one PR:

1. **Wrapper-side rails** (commits f93a0be -> 7e481f3) -- macOS Electron stubs plus seven phases of "rails not barriers" guards on the manual permission flow.
2. **Recovery infrastructure** (commit 7be25b6) -- explicit update flow + COMPAT.md tested-versions matrix + launcher warnings on untested asar versions.

**Rails, not barriers** -- the user is inside the threat model with their own agent. The wrapper provides defaults that constrain casual exploration without fighting the user's intent. Time and space caps protect against drift in upstream defaults, not against the user, and not against a sophisticated attacker.

### Wrapper rails (phases 1-7)

- **Phase 1** -- remove the 12 `LocalAgentModeSessions_$_*Bridge*` IPC overrides so the asar's `LocalAgentModeSessionManager` can drive the manual user-acceptance flow (Allow/Deny clicks finally reach the bridge).
- **Phase 2** -- TTL cap on `autoPermissionsModeEnabled` / `bypassPermissionsModeEnabled` at 60 min wall-clock (CLOCK_MONOTONIC via libuv). Closure-private state, no persistent disk state. Re-toggling resets the deadline; toggling off clears it.
- **Phase 3** -- `createMountSymlinks` refuses mount roots at `$HOME`, `/`, or anything shallower than three path segments. Principle, not enumeration: no allowlist of sensitive paths.
- **Phase 4** -- opt-in canary on `ipc_tap` (`CLAUDE_COWORK_IPC_TAP=1` or `CLAUDE_COWORK_BRIDGE_CANARY=1`) that warns once when a previously-unseen bridge-related channel suffix registers. Regression alarm for "Anthropic shipped a new bridge channel between releases."
- **Phase 5** (predicate only; wiring deferred) -- pure path-policy module: `tool in READ_SET` AND every path arg, `realpath`-resolved, is contained in `realpath(cwd)`. READ_SET = `{read_file, list_dir, grep, head, tail}`. Wiring blocked on a live IPC trace of the renderer-side auto-approve seam.
- **Phase 6** (predicate only; wiring deferred) -- audit of bundled `index.js` enumerated 11 mutating + 8 read-only scheduled-task channels across two namespaces. Predicate and refused-handler factory ship now.
- **Phase 7** -- startup self-test runs once per `app.whenReady()`. Writes one JSON line to `~/.local/state/claude-cowork/logs/cowork-startup.log` confirming the Phase 1-3 rails are engaged.

### Recovery (new in this PR)

- **`COMPAT.md`** at repo root with a machine-readable `LAST_TESTED_ASAR_VERSION` line plus a human-readable status matrix. Bumped via PR each time a contributor confirms a new asar works end-to-end. Status indicators are ASCII (`[OK]`, `[PARTIAL]`, `[FAIL]`, `[UNTESTED]`).
- **`install.sh`** stops auto-downloading silently. New prompt shows the last-tested vs. latest-CDN versions and asks for confirmation before download. Refuses piped (non-interactive) installs without `--force`. `--update` flag re-fetches / re-extracts / re-stubs / re-patches without rebuilding the launcher.
- **Launcher (`claude-desktop`, `claude-cowork`)** reads a version sentinel written at install time and prints a one-time `[WARN]` line to stderr if the installed asar is newer than the last tested. Sentinel at `$LOG_DIR/.last-warned-asar-version` suppresses re-warning for the same version.
- **`doctor`** reports the installed-vs-tested version state as an `[OK]` or `[WARN]` line.
- **`README.md`** has a new **Recovery** section walking users through rollback when an update breaks their install.

## Files changed

- `stubs/cowork/ipc_overrides.js` -- bridge handlers removed (Phase 1).
- `stubs/cowork/auto_permissions_cap.js` (new) -- Phase 2 module.
- `stubs/cowork/mount_root_bound.js` (new) -- Phase 3 predicate.
- `stubs/cowork/bridge_canary.js` (new) -- Phase 4 module.
- `stubs/cowork/tool_preflight_policy.js` (new) -- Phase 5 predicate.
- `stubs/cowork/scheduled_task_gate.js` (new) -- Phase 6 predicate.
- `stubs/cowork/startup_check.js` (new) -- Phase 7 module.
- `stubs/frame-fix/frame-fix-wrapper.js` -- wires Phases 2, 4, 7; activity-stubs base.
- `stubs/@ant/claude-swift/js/index.js` -- wires Phase 3.
- `COMPAT.md` (new) -- tested-versions matrix.
- `install.sh` -- compat helpers, prompt-gated download, `--update` flag, sentinel seeding, doctor version check, updated launcher heredoc.
- `README.md` -- Quick Start prompt note, Updating subsection, Recovery section.
- `tests/node/current-path/*.test.cjs` -- 6 new test files for the new modules, plus rewrites of bridge tests.

Total: ~1700 LOC, 461 tests pass (was 403 -- +58 new). All commits squash cleanly.

## Discovery blockers (Phases 5, 6)

Phases 5 and 6 ship predicates only. Wiring requires live IPC traces this development environment cannot produce.

- **Phase 5** -- `autoPermissionsModeEnabled` appears only in preference schemas in the bundled `index.js`; the actual auto-approve decision lives in the renderer. Wiring requires `CLAUDE_COWORK_IPC_TAP=1` against a real Cowork session that triggers a tool preflight, and the resulting channel suffix recorded.
- **Phase 6** -- 11 mutating scheduled-task channels enumerated; bridge-reachability cannot be determined from static grep. Resolution requires a live trace of bridge messages exercising scheduled-task creation/modification.

Per the spec, the predicates ship now (pure, tested, correct); the wiring lands in follow-up PRs after live traces.

## Coordination

This PR's first commit (activity stubs) is duplicated by #103, #110, #115, #116 -- all closable after merge. Phase 1 (bridge override removal) overlaps with #113's same change; the second to land is a no-op for that file. Everything else is disjoint from #113 (which adds feature-coverage and PKGBUILD changes orthogonal to the rails + recovery scope).

OAuth-completion leg of #104 remains with #100. node-pty Linux ELF remains with #109.

## Verification

- `node --test tests/node/current-path/*.test.cjs` -- **461 pass, 0 fail** (was 403; +58 new)
- `node --check` on every modified `.js`
- `bash -n install.sh` -- parses clean after recovery commit
- Smoke test of `compat_version_is_newer` across `>`, `<`, `==`, and the lexical-fail case (1.5.10 > 1.5.9). All correct.
- Smoke test of launcher `_warn_if_untested_asar` across all five paths: fires once, then silent, fires again after sentinel cleared, silent on same-version, silent on older-installed.
- `LC_ALL=C grep -P '[^\x00-\x7F]'` confirms `COMPAT.md` is ASCII and new blocks in `install.sh` and `README.md` are ASCII.
- **Not** yet smoke-tested on a live Linux desktop (maintainer's local environment is unavailable). End-to-end install/--update/launch verification deferred to a contributor with a working desktop.

## Test plan

- [x] Full Node test suite (461 pass)
- [x] Bash syntax check on install.sh
- [x] Helper smoke tests (compat_version_is_newer, launcher warn-once)
- [x] ASCII audit on new content
- [ ] `./install.sh --doctor` against a fresh install -- expect new `[OK] Asar version` line
- [ ] `./install.sh` (interactive) on a fresh box -- expect prompt before download
- [ ] `./install.sh` (piped, no --force) -- expect refusal
- [ ] `claude-desktop --update` on an existing install -- expect prompt, then re-extract+repack
- [ ] Launch on a newer-than-tested asar -- expect `[WARN]` line once, then silent
- [ ] Sign in via Google -- `claude://` redirect re-focuses without crash (covered by Phase 0 stubs)

Closes #106. Closes the crash-dialog leg of #104. Addresses majority of #107 with the rails approach plus the user-facing escape hatch in COMPAT.md. The OAuth-completion leg of #104 remains with PR #100.

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -1,0 +1,68 @@
+# Compatibility
+
+This file tracks which Claude Desktop asar versions have been verified
+to work with claude-cowork-linux. `install.sh` and the launcher grep the
+machine-readable lines below; the table further down is for humans.
+
+<!-- machine-readable; do not remove the next two lines -->
+<!-- LAST_TESTED_ASAR_VERSION=1.6259.1 -->
+<!-- LAST_TESTED_DATE=2026-05-14 -->
+
+## Tested versions
+
+| Asar     | Status     | Date       | Notes                                                |
+|:---------|:-----------|:-----------|:-----------------------------------------------------|
+| 1.6259.1 | [OK]       | 2026-05-14 | v5.1.0 baseline. Activity stubs and bridge rails verified via test suite. |
+| 1.6608.2 | [PARTIAL]  | 2026-05-07 | `/setup-cowork` reports "Unsupported platform: linux-x64" -- see issue #114. |
+| 1.6700.0 | [UNTESTED] | -          | Not yet exercised by any contributor.                |
+
+Status legend:
+
+- `[OK]`       -- end-to-end exercised, ships clean.
+- `[PARTIAL]`  -- launches and most features work, but at least one known regression. See Notes.
+- `[FAIL]`     -- known to be broken; do not update to this version.
+- `[UNTESTED]` -- no contributor has reported results.
+
+## Reporting a tested version
+
+Open a PR that bumps `LAST_TESTED_ASAR_VERSION` (the HTML comment line
+above) and adds a row to the table. In the PR body, include:
+
+1. Distro and desktop environment (e.g. "Arch Linux + Hyprland (Wayland)").
+2. Electron version (`electron --version` from your install).
+3. Which features you exercised:
+   - First-run install via `install.sh`
+   - Login via the `claude://` OAuth callback
+   - Cowork session start with a tool permission prompt
+   - In-app shell / PTY panel
+   - At least one MCP tool call
+
+Reports that only confirm "the app launches" are useful but should be
+flagged with `[PARTIAL]` until a fuller exercise is recorded.
+
+## Why this file exists
+
+`install.sh` no longer auto-downloads the latest Claude Desktop asar
+without prompting. The installer reads `LAST_TESTED_ASAR_VERSION` from
+this file, fetches the latest version available on Anthropic's CDN, and
+warns if the latest is newer than the last tested. The user decides
+whether to proceed.
+
+The launcher (`claude-desktop`, `claude-cowork`) reads the installed
+asar version from `$INSTALL_DIR/.installed-asar-version` (written by
+`install.sh` at the end of install/update) and prints a `[WARN]` line
+to stderr if the installed version is newer than the last tested. The
+warning fires once per version change -- after the user sees it once,
+they will not see it again for the same installed asar.
+
+To re-trigger the warning manually, delete
+`$XDG_STATE_HOME/claude-cowork/logs/.last-warned-asar-version`.
+
+## See also
+
+- README.md "Recovery" section for what to do if your install breaks
+  after an update.
+- `install.sh --doctor` reports the installed-vs-tested version state
+  as an `[OK]` or `[WARN]` line.
+- `claude-desktop --update` re-runs the download/extract/repack flow
+  with the same prompt as a fresh install.
--- a/README.md
+++ b/README.md
@@ -86,14 +86,16 @@ Run `./install.sh --doctor` (or `claude-desktop --doctor`) after install to vali
 ```bash
 git clone https://github.com/johnzfitch/claude-cowork-linux.git
 cd claude-cowork-linux
-./install.sh          # auto-downloads the latest DMG via Node.js
+./install.sh          # prompts before downloading the Claude Desktop asar
 claude-desktop
 ```
+
+The installer prompts before downloading the Claude Desktop asar so you can confirm the version. See [COMPAT.md](COMPAT.md) for the tested-versions matrix. Pass `--force` (or `--yes`) to skip the prompt for scripted installs.
 
 ### Method 2: AUR (Arch Linux)
 
 ```bash
-yay -S claude-cowork-linux       # auto-downloads the latest DMG
+yay -S claude-cowork-linux
 ```
 
 ### Method 3: curl pipe
@@ -102,13 +104,27 @@ yay -S claude-cowork-linux       # auto-downloads the latest DMG
 bash <(curl -fsSL https://raw.githubusercontent.com/johnzfitch/claude-cowork-linux/master/install.sh)
 ```
 
-The installer automatically downloads the latest Claude Desktop DMG using Node.js (`scripts/fetch-dmg.js`). You can also provide a DMG manually:
+Piped (non-interactive) installs refuse to download without an explicit confirmation flag. Use:
+
+```bash
+bash <(curl -fsSL .../install.sh) --force
+```
+
+You can also provide an archive manually:
 
 ```bash
 ./install.sh ~/Downloads/Claude-*.dmg
 # or
-CLAUDE_DMG=~/Downloads/Claude-1.1.4010.dmg ./install.sh
+CLAUDE_ARCHIVE=~/Downloads/Claude-1.6259.1.dmg ./install.sh
 ```
+
+### Updating
+
+```bash
+claude-desktop --update          # re-fetch and repack with prompt
+```
+
+The launcher prints a `[WARN]` line once if your installed asar is newer than the last tested version listed in [COMPAT.md](COMPAT.md). To dismiss the warning permanently for the current version, just keep using the app -- it only fires once per asar version change.
 
 > [!IMPORTANT]
 > This repo does not include Anthropic's proprietary code. The installer downloads it directly from Anthropic's CDN.
@@ -421,6 +437,68 @@ npm install -g electron @electron/asar
 ```
 
 </details>
+
+---
+
+## Recovery
+
+If your install was working and then stopped after an update, the upstream
+asar likely changed in a way claude-cowork-linux has not yet caught up to.
+Two recovery paths:
+
+### Option 1: roll back to the last tested asar
+
+The installer reads `LAST_TESTED_ASAR_VERSION` from [COMPAT.md](COMPAT.md)
+and prompts before downloading. To roll back:
+
+```bash
+cd ~/.local/share/claude-desktop
+git pull
+
+# Wipe cached chromium artifacts (preserves your sessions and credentials):
+rm -rf ~/.config/Claude/Cache \
+       "~/.config/Claude/Code Cache" \
+       ~/.config/Claude/GPUCache \
+       ~/.config/Claude/DawnGraphiteCache
+
+# Reinstall:
+bash install.sh --force
+```
+
+When the prompt offers to download an untested version, decline and supply
+the last-tested archive manually:
+
+```bash
+CLAUDE_ARCHIVE=/path/to/Claude-<last-tested>.dmg bash install.sh --force
+```
+
+### Option 2: stay on your current install, wait for a fix
+
+The launcher prints a `[WARN]` line on the first launch after an untested
+version is installed. If the warning appears and the app is broken for
+you, the issue is likely already being tracked. Search the open issues at
+<https://github.com/johnzfitch/claude-cowork-linux/issues> for the asar
+version in the warning.
+
+To update later once a fix is published:
+
+```bash
+claude-desktop --update
+```
+
+You will be prompted before the new asar is downloaded.
+
+### Diagnostics
+
+```bash
+claude-desktop --doctor
+```
+
+Reports the installed-vs-tested asar versions plus the rest of the
+preflight checks. An `[OK] Asar version: X (matches last tested)` line
+means you are on the verified baseline. A `[WARN] Asar version: X (newer
+than last tested Y -- see COMPAT.md)` line means you are ahead of the
+matrix.
 
 ---
 
--- a/install.sh
+++ b/install.sh
@@ -85,6 +85,46 @@ detect_pkg_manager() {
 }
 
 # ============================================================
+# Compatibility helpers (COMPAT.md)
+# ============================================================
+#
+# install.sh and the launcher grep machine-readable lines from
+# COMPAT.md to find the last tested asar version. The user chooses
+# whether to install a newer (untested) version when prompted.
+
+compat_read_last_tested() {
+    # Reads LAST_TESTED_ASAR_VERSION from the given COMPAT.md (or the
+    # one next to this script if no argument given). Falls back to a
+    # baked-in default if the file is missing.
+    local compat_file="${1:-}"
+    if [[ -z "$compat_file" ]]; then
+        compat_file="$(dirname "$0")/COMPAT.md"
+    fi
+    if [[ -f "$compat_file" ]]; then
+        local value
+        value=$(grep -oE '^<!-- LAST_TESTED_ASAR_VERSION=[^[:space:]]+ -->$' "$compat_file" \
+            | head -1 | sed -E 's/.*=([^[:space:]]+) -->/\1/')
+        if [[ -n "$value" ]]; then
+            printf '%s' "$value"
+            return 0
+        fi
+    fi
+    # Baked-in fallback so a malformed/missing COMPAT.md never blocks install.
+    printf '1.6259.1'
+}
+
+compat_version_is_newer() {
+    # Returns 0 (true) if $1 > $2 by semver-ish ordering (sort -V).
+    # Returns 1 otherwise (equal or older).
+    local a="$1" b="$2"
+    [[ -z "$a" || -z "$b" ]] && return 1
+    [[ "$a" == "$b" ]] && return 1
+    local newer
+    newer=$(printf '%s\n%s\n' "$a" "$b" | sort -V | tail -n1)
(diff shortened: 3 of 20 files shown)


michael-greider:
let me cook first!! I categorized and am working out the resulting bugs from the core issue . Dialed it in to 5 categories- want you to see it first

johnzfitch:
Haha okay cook it up my dude ^^

Thank you!

michael-greider:
I've almost got things buttoned up here my prediction in #107 is looking like it's the case. The last update broke the application, but was inherently a better approach than my #98. I'm running some tests on my main driver and my sandbox so I can see if there's any caching that pins regressions even if you remove the package and reinstall vs a clean install on a computer I'm willing to bomb electron between every test. 

As it stands from what I've seen this fix shows UI features that did not appear before and by extension breaks certain command calls. For example a per file allow list for permission requests shows up that never did before, exactly why your approach was better- with the caveat of certain grep commands and the rm command not working. It's tricky but I almost have it wrapped up, had to (NSFW warning) write most of it by hand as Claude code is not very good at incremental binary parsing. Out of the 35 skipped calls from #105 they expand to ~108 minor path resolution errors especially with any MCPs- but they can be chopped down with 5-6 underlying errors-- will drop a report soon. 

I'm salty #105 was better so I'm shooting to also cover the 7 Mac OS only calls that slip in with the approach

johnzfitch:
Sorry for the delay. It’s been difficult trying to expand open the bridge whilst narrowing blast radius on Linux. The asar really just lets the model do whatever because it assumes windows/Mac will prevent it from causing permanent damage. Linux doesn’t have such protections so I am thinking of a genuinely new permissions scheme that tries to prevent whilst also allows for it to work as users would like. Without frustration/annoysnce. It’s delicate. Would appreciate more perspectives here.

johnzfitch:
Throwing in a perspective since I've been deep in this for the rails work. The catch-22 dissolves if you flip the framing: Linux is the OS with the **best** sandbox primitives for this, not the worst. macOS App Sandbox and Windows AppContainer are heavy, opaque, and policy-locked by the OS vendor. `bwrap` (already a dep) is fine-grained, declarative, and lets you say exactly what the agent can touch with **kernel** enforcement — not JS prompts.

Per-tool prompt-or-allow at the asar layer is the design that produces frustration AND the design that lets a slip cause permanent damage, because by the time the prompt fires the model is moments away from the destructive action. The fatigue and the danger are the same problem.

The shape I'd suggest: **session-level consent**, a few knobs set once at session start, kernel-enforced thereafter.

1. **Filesystem boundary = cwd.** Already enforced lexically by Phase 3. Add a bwrap layer that bind-mounts `/work=cwd` writable and everything else read-only or invisible. Writes outside cwd return EROFS at the kernel level — there's no prompt because there's nothing to prompt about. Eliminates ~70% of "is this safe?" decisions.
2. **Network policy**, one of: `none` / `claude-only` (allowlist `*.anthropic.com` + bridge via `session.webRequest`) / `open`. Implemented at `--unshare-net` + forwarded socket, or shared-net + origin allowlist. Per-tool network prompts go away.
3. **Exec policy**, one of: `read-only-tools` (Phase 5 READ_SET only; no `bash`, no `find -exec`) / `agent-tools-with-confirm` (current behavior for the destructive ones) / `full`.

The user makes a 3-knob deal at session start. Inside those bounds, auto-mode stops being scary because the *kernel* bounds the blast. The bridge stays open. `tool_permission_request` still surfaces for the genuinely interactive cases. Prompt fatigue goes away because 80% of tool calls are `read_file` / `list_dir` / `grep` that are inherently safe when fs is bound to cwd.

The rails already in this PR are the JS-layer half of this:
- Phase 2: time bound (TTL cap on the toggle)
- Phase 3: space bound (cwd predicate on mounts)
- Phase 5: operation bound (READ_SET predicate)
- Phase 7: startup sanity-check that all three are engaged

The other half is `bwrap` + `webRequest`. Both already in the project's toolbox. The "genuinely new permissions scheme" is mostly **removing** JS prompts, not adding them — let the kernel enforce the bounds the user already agreed to at session start.

Concrete next step that would inform the design: pick one knob (writes outside cwd) and prototype the bwrap profile that makes it impossible at the kernel layer. If it works for that one, the same shape scales to the other two. ~100-line PR, not a redesign. Happy to take a swing if useful.

(Re Phase 5 wiring on this branch: the channels are discoverable from the bundle but the auto-approve decision is renderer-side, so the cleanest seam is to wrap inbound `respondToToolPermission`, correlate by `requestId`, evaluate the predicate, and revert the toggle on the first slip. Currently shipped as predicate-only; happy to wire it once you're ready.)

---

johnzfitch:
Yeah but this assumes the kernel is safe. Copy fail isn’t a one off, its a whole new category of exploit. All the agent needs is a writeable directory to expand permission scope. That’s why time is the only method of scoping I can currently come up with for auto mode. It turns auto forever, infinite blast radius to auto for a time, set by the user, capped at an hour. Whether that’s enough time to prevent a misaligned agent is another question altogether.

additionally, time is easily manipulated on system. So something like this has to be focused around an unmanipulated vector that maps to time. Say tool calls, or some sort of heart beat like system within the Asar itself. Something that would break the bridge if it were manipulated or dropped.

johnzfitch:
Fair pushback on both fronts. Let me concede where you're right and push back where I think the picture is rosier than you're framing.

**On kernel-trust**: you're right that bwrap isn't a ceiling. The agent-as-confused-deputy is the obvious failure: bind `/` read-only with `/work` writable, and the agent asks the wrapper to copy `/etc/passwd` into `/work/foo` and reads it. The ro-bind didn't help — read access to mounted secrets is still read access. Add `copy_file_range` bugs, setuid binaries surviving in the namespace, /proc race conditions, user-namespace CVEs. So "bwrap as the answer" is too strong. The honest framing is "kernel-ns raises the floor for cost-of-exploit, doesn't put a ceiling on."

**On time being manipulable**: actually less so than you might think for Phase 2's specific implementation. `setTimeout` in Node uses libuv's monotonic clock (`uv__hrtime` → `clock_gettime(CLOCK_MONOTONIC)` on Linux), not `Date.now()`. CLOCK_MONOTONIC isn't settable from userspace without CAP_SYS_TIME — even root in a user namespace can't move it backward. So clock-manipulation attacks against the existing cap require:

1. Running with CAP_SYS_TIME outside the namespace (separate threat, much harder)
2. Restarting the wrapper process with pre-toggled config — but the toggle isn't persisted to disk, it's runtime state in the asar's prefs cache, so a fresh process defaults to off
3. /proc/<pid>/mem patching of libuv internals — needs CAP_SYS_PTRACE

That's a higher bar than "time is easily manipulated" implies. The Phase 2 cap is honestly OK against time manipulation by itself. What it's not OK against is *coverage* — a single monotonic vector is brittle as the only line.

**Where I think you're directionally correct**: you want multiple monotonic bounds, each tamper-resistant in a different physical quantity. The agent has to beat *all of them* to keep auto-mode permissive past the safety horizon. Three I'd suggest stacking with the existing wall-clock cap:

1. **Tool-call counter.** Increment in `handleToolPermission` (we just traced this for Phase 5 — every tool call funnels through it). Cap at N calls. Monotonic in *agent activity*, not wall time. Agent can't accelerate decay (calls cost real work). Agent can't decelerate it either (idle agent = no calls = no permissions used anyway).
2. **Bridge heartbeat liveness.** If the bridge is open (your direction here), it likely already maintains the WebSocket. Wrapper observes the ack cadence; if it stops for >T seconds, revert. Ties enforcement to a signal Anthropic controls, not local agent. Bonus: the same condition that increases danger (broken bridge during an agent action) also engages the safety. (I don't actually know if the asar exposes heartbeat observability cleanly — needs a tap session to confirm before promising the wiring.)
3. **Read-set inversion as a forcing function for the cap.** The existing READ_SET predicate (Phase 5) refuses auto-approval for anything outside cwd or anything write/exec. So the only thing the cap protects is *reading inside cwd* — a much smaller blast radius than "anything." Time becomes a bound on *how long the read window stays open*, not "how long until permanent damage." That's a more defensible scoping problem.

Conjunction is much harder to defeat than any single primitive:

- wall-clock cap (CLOCK_MONOTONIC) — resists Date.now manipulation
- tool-call cap — resists clock manipulation entirely
- heartbeat revert — resists local agent forgery
- cwd predicate — orthogonal: bounds space, not time

For the agent to keep auto-mode permissive past the safety horizon, it has to clear all four. Each is monotonic in a different physical quantity (time, agent activity, network liveness, filesystem geometry).

**Concrete offer**: I can extend Phase 2 to add the tool-call counter cleanly — same module, ~50 LOC + tests, fires whichever cap hits first. The heartbeat hook is a sketch until we trace the bridge's WebSocket ack pattern in a real session. The bwrap layer goes from "the answer" to "one of four bounds" — useful for the categories it does handle (writes outside cwd are genuinely blocked at the kernel layer), honest about the categories it doesn't (mounted-secret reads).

Want me to ship the tool-call counter as Phase 2b on this branch?

---

johnzfitch:
There we go, that’s a much better plan. Although I do think an additional layer needs to be considered on the tool count. Can we valid count to valid SHA ID or whatever marries time + tool details on Anthropic‘s end that persists whenever a conversation is loaded from restart?

That is, it needs to be checked by something bound to the session and cross referenced with a non-local and local primitive. Otherwise an agent can just reset the counter locally or attempt to overflow the pipe.

PR #109, opened 2026-05-08
fix(node-pty): vendor Linux ELF pty.node, copy on launch

wlcarden:
## Summary

The DMG ships `node-pty`'s `pty.node` as a macOS Mach-O universal binary (x86_64 + arm64). On Linux, `dlopen` rejects it (`invalid ELF header`); node-pty's loader falls through to a non-existent `Debug` build and re-throws the original error. Any feature calling `LocalSessions.startShellPty` (in-app terminal panel, shell tool execution) fails. Cowork's CLI spawning uses `vm.spawn()` rather than pty, so cowork itself was unaffected — but the in-app shell is broken on every launch.

## Root cause

```
Failed to load node-pty {
  error: Error: <tmp-path>: invalid ELF header
      at process.func [as dlopen] (node:electron/js2c/node_init:2:2617)
      ...
  code: 'ERR_DLOPEN_FAILED'
}
```

`linux-app-extracted/node_modules/node-pty/build/Release/pty.node` reports as `Mach-O universal binary with 2 architectures`. Inspection of `node-pty/binding.gyp` confirms `spawn-helper` is gated to `OS=="mac"` only, and `pty.cc` only references `helper_path` inside `#if defined(__APPLE__)` — so on Linux just the `pty.node` ELF is needed (Linux uses `forkpty()` directly).

## Fix

Vendor a Linux ELF build of `node-pty@1.1.0-beta34` at `stubs/node-pty-linux/pty.node` (75 856 bytes, built against Electron 41 ABI: `target=41.5.0 runtime=electron disturl=electronjs.org/headers`). `launch.sh` copies it into `linux-app-extracted/node_modules/node-pty/build/Release/pty.node` on each repack, mirroring the existing pattern for `stubs/cowork/*.js` and `stubs/frame-fix/*.js`.

This is the simplest fix and follows the repo's "stage Linux replacements outside the extracted tree, copy on launch" idiom. The tradeoff is a 76 KB native binary in git that must be re-vendored on Electron major-version bumps.

### Alternative considered

A `scripts/build-pty.sh` invoked from `launch.sh` could rebuild on each user's machine — no binary in git, but adds `node`, `node-gyp`, `python3`, `make`, `g++` as install-time dependencies and risks per-distro build failures. Happy to switch to this approach if preferred.

## Test plan

- [x] `file stubs/node-pty-linux/pty.node` reports `ELF 64-bit LSB shared object, x86-64`
- [x] After `launch.sh` repacks, the asar-internal `pty.node` is the Linux ELF (verify with `asar extract-file`)
- [x] No `Cannot find module '../build/Debug/pty.node'` error in `~/.local/state/claude-cowork/logs/startup.log`
- [ ] In-app terminal / shell-PTY panel opens without error
- [ ] Cowork sessions still work (regression check — they don't use pty)


johnzfitch:
Still relevant for in-app shell/PTY. Needs rebase onto current master. The ABI concern (Electron 41 binary vs users on Electron 42+) is worth noting in the PR -- consider adding a check in launch.sh that warns if the vendored ELF was built against a different Electron major.

---

johnzfitch:
Closing as superseded by #127, now merged to `master`. #127 explicitly consolidates this PR — `launch.sh` now copies the Linux ELF `pty.node` into the extracted tree on each repack when present, which is the node-pty fix from here.

---

PR #111, opened 2026-05-08
fix(ipc-overrides): remove bridge handler overrides blocking dispatch

wlcarden:
## Summary

`stubs/cowork/ipc_overrides.js` registers 12 no-op overrides for `LocalAgentModeSessions_$_*Bridge*` channels. These were originally added defensively to prevent `No handler registered` warnings, but they intercept the asar's own `LocalAgentModeSessionManager` registration and replace its dispatch handlers with stubs that always return `null` / `disconnected` / `consented:false`. The asar's WebSocket bridge to `wss://bridge.claudeusercontent.com` therefore never connects, and dispatch reports "can't connect" for all users.

## Root cause

The handlers being shadowed are managed internally by the asar's `LocalAgentModeSessionManager` — it registers them via `ipcMain.handle()` to drive its WebSocket bridge. Our `matchOverride()` in the registration interceptor finds our entries first and substitutes them, so the asar's real handlers never run.

## Fix

Remove all 12 bridge handler entries:

- `abandonBridgeEnvironment`
- `deleteBridgeAgentMemory`
- `deleteBridgeSession`
- `getBridgeConsent`
- `getSessionsBridgeEnabled`
- `kickBridgePoll`
- `onBridgePermissionPreflight`
- `resetBridge`
- `resetBridgeSession`
- `respondBridgePermissionPreflight`
- `sessionsBridgeStatus`
- `setSessionsBridgeEnabled`

A comment block remains at the deletion site explaining why these channels must not be overridden.

## Test changes

`tests/node/current-path/security_hardening.test.cjs` had three tests asserting the broken behavior (e.g. `getBridgeConsent returns consented: false` — i.e. the override returns `false`). These are rewritten to lock in the correct behavior: `matchOverride()` for any of the 12 bridge channels must return falsy, so the asar's handler runs unimpeded.

`tests/node/current-path/ipc_overrides.test.cjs` similarly drops the override-behavior tests for these channels.

## Test plan

- [x] `node --test tests/node/current-path/*.test.cjs` — 98 affected tests pass, none regressed
- [x] `node --check stubs/cowork/ipc_overrides.js`
- [ ] Verify dispatch connects: `[sessions-bridge] WebSocket connected to wss://bridge.claudeusercontent.com` appears in startup log
- [ ] Verify "Polling for work" loop runs and tasks are received

### DIFF
--- a/stubs/cowork/ipc_overrides.js
+++ b/stubs/cowork/ipc_overrides.js
@@ -387,66 +387,12 @@ function createOverrideRegistry(getProcessState) {
     },
 
     // ================================================================
-    // LocalAgentModeSessions — Dispatch/Bridge handlers
+    // Bridge handlers REMOVED: these no-op overrides blocked the asar's own
+    // LocalAgentModeSessionManager from establishing the WebSocket bridge
+    // to wss://bridge.claudeusercontent.com. The asar registers and manages
+    // these handlers internally; with our overrides intercepting registration,
+    // the bridge never connected and dispatch reported "can't connect."
     // ================================================================
-
-    'LocalAgentModeSessions_$_abandonBridgeEnvironment': async (_event, ...args) => {
-      vlog('[ipc:abandonBridgeEnvironment] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_deleteBridgeAgentMemory': async (_event, ...args) => {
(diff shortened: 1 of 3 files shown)


johnzfitch:
Subsumed by #108 (v5.1.0) Phase 1 -- same twelve bridge handlers removed. Thanks for the careful writeup.

---

PR #110, opened 2026-05-08
fix for invalidateCurrentActivity is not a function

Asitis:
## What does this change?

Fixes https://github.com/johnzfitch/claude-cowork-linux/issues/106

## Type
- [x] Bug fix

## Testing

- Distro / desktop: Ubuntu 26.04 LTS
- Tested with `./install.sh --doctor`: Yes
- Tested a full Cowork session: Yes

## Security impact
- [X] This change does touch security-sensitive code

## Checklist
- [X] No API keys, tokens, `.env` files, or log output with credentials included
- [X] Commit messages follow the project style (no emoji, brief summary + explanation)
- [X] `./install.sh --doctor` passes cleanly on my system

PR #113, opened 2026-05-08
Linux port coverage to incorporate #105 

### LINKED ISSUE #106: [BUG] TypeError: CA.app.invalidateCurrentActivity is not a function on startup

michael-greider:
The predictions i had were exactly right, #105 was indeed a superior approach to #98 -- with the caveat of needing a LOT of manual path translations. 
commitpic2
commitpic1

^^as you can see here these are just 2 mini-features that did not previously show up in the UI (ignore my claude preference being set to "smooth 90s black guy who doesn't play with no jive turkey")
                                                                                     
 11 commits across 5 files (PKGBUILD, enable-cowork.py, 3 stubs). I will have more on the way depending on what other things people can come up with. Here is a graphical representation of where the project stands with #105 and this one: 
commitexample

                                    
                                                            
What was changed:                                                                   
  - `enable-cowork.py`: fingerprint-based getHostPlatform patch (replaces global-sub bug — hits only   
  the unique darwin-x64/win32-x64 body, not all 3 occurrences)                                         
  - `claude-swift/index.js`: complete computerUse namespace (apps + display sub-namespaces — asar
  accesses `computerUse.apps.listInstalled()` not flat methods), updater.getShipItJobInfo, vm guest
  request methods for cliPluginBridge, darwin-arch SDK URL
  - `ipc_overrides.js`: removed bridge handler overrides per PR #111 — lets the asar drive WebSocket
  dispatch directly
  - `PKGBUILD`: app.asar inside resources/ to mirror macOS Contents/Resources layout, launcher cache
  hygiene

Note: the computer use feature is one of our biggest gaps and i didn't wanna touch it without @johnzfitch deciding how we'd all go about it. The native macos binary has of course system and accessibility settings that will open the mac settings app. I say keep it simple and just map the actual equivalent linux deps/packages that will satisfy the program's needs and make a qtquick widget that will handle the linkage/stubbing. BUT the fact it even requests for it now goes to show why #105 was the right direction with more foot work (i spent many hours on this because i was salty it is in fact better than #98) 

Subsumes / extends:
  - Extends #108 (macOS Electron stubs base)
  - Subsumes #103 (app activity APIs)
  - Subsumes #110 (invalidateCurrentActivity)
  - Subsumes #111 (bridge handler overrides)
  - Subsumes #112 (claude-swift guest methods)
  - Addresses majority of #107 (active darwin-x64 path gates)
  - Fixes #106 (invalidateCurrentActivity)

Todo fixes-- might also be done already, there is gonna have to be significant stability leg work
I would say merge minimal working fixes or we all just do the leg work to make sure all paths resolve
my assumption in #107 was on the money. You fix 10 path resolutions > find 5 more > fix > find 2 < fix just those 2 > the rest resolve without having to alter the code as significantly. With so many path resolutions thing's can be MUCH more systematic and updates from Anthropic will take 5 minutes to handle if they break things. 

  - #100 / #104 OAuth callback (install.sh-flow specific; AUR PKGBUILD uses xdg-mime registration —
  different code path)
  - #109 node-pty (handled in outer AUR PKGBUILD via `npm install node-pty + npx @electron/rebuild`,
  different mechanism than the vendored ELF approach)
  - ~10 low-impact `~/Library/` path leaks from #107 (NativeMessagingHost, VS Code project detection —
  feature-degraded harmless on Linux) 

The biggest problem i had was- 
  The disclaimer-arg redirect (commit 11) fixes `mcp__workspace__bash` auto-denial: the asar passes
  `claude-code/<ver>/claude.app/Contents/MacOS/claude` as the SDK exec arg; falling back to
  /usr/bin/claude (or any non-version-matched binary) silently breaks the IPC handshake for
  tool_permission_request, so workspace tools deny without ever prompting the UI. Now resolves to
  ~/.config/Claude/claude-code-vm/<ver>/claude.

### DIFF
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -207,12 +207,14 @@ JSEOF
 package() {
     cd "${srcdir}"
 
-    # Install repacked app.asar
+    # Mirror macOS Contents/Resources/ layout. Two lookups in the asar key off
+    # this layout: path.join(resourcesPath, 'app.asar', ...) for MCP nodeHost
+    # discovery, and path.join(path.dirname(resourcesPath), 'Helpers',
+    # 'disclaimer') for the disclaimer wrapper. Putting app.asar inside
+    # resources/ and the disclaimer one level up makes both resolve.
     install -Dm644 "${srcdir}/app.asar" \
-                   "${pkgdir}/usr/lib/claude-cowork/app.asar"
+                   "${pkgdir}/usr/lib/claude-cowork/resources/app.asar"
 
-    # Resources in our namespace (avoid /usr/lib/electronNN/, foreign-owned).
-    # i18n in BOTH root and i18n/ -- bundle reads from both paths.
     install -d "${pkgdir}/usr/lib/claude-cowork/resources/i18n"
     install -m644 "${srcdir}/linux-app-extracted/resources/"*.json \
         "${pkgdir}/usr/lib/claude-cowork/resources/"
@@ -277,7 +279,7 @@ elif [[ -f /proc/sys/user/max_user_namespaces ]]; then
     [[ "$_max_ns" -gt 0 ]] 2>/dev/null && _sandbox_flag=""
 fi
 
-exec electron /usr/lib/claude-cowork/app.asar \
+exec electron /usr/lib/claude-cowork/resources/app.asar \
     ${_sandbox_flag} \
     --disable-gpu \
     --class=Claude \
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -250,6 +250,17 @@ EOF
 #!/bin/bash
 # Claude Cowork Linux launcher
 
+# Wipe build-time caches so asar patches and chromium GPU/font caches
+# rebuild fresh. User data (IndexedDB, Local Storage, Cookies, sessions)
+# is left intact.
+_CFG="${XDG_CONFIG_HOME:-$HOME/.config}/Claude"
+for _c in "$_CFG/Cache" "$_CFG/Code Cache" "$_CFG/GPUCache" "$_CFG/CachedData" \
+          "$_CFG/DawnGraphiteCache" "$_CFG/DawnWebGPUCache" "$_CFG/fcache" \
+          "$_CFG/Shared Dictionary/cache"; do
+    [ -e "$_c" ] && rm -rf "$_c"
+done
+unset _CFG _c
+
 if [[ -n "$WAYLAND_DISPLAY" ]] || [[ "$XDG_SESSION_TYPE" == "wayland" ]]; then
     export ELECTRON_OZONE_PLATFORM_HINT=wayland
 fi
--- a/enable-cowork.py
+++ b/enable-cowork.py
@@ -100,32 +100,36 @@ def patch_file(filepath):
     print(f"  {func_name}() now returns {{status:\"supported\"}} unconditionally")
     return True
 
+HOST_PLATFORM_FINGERPRINT_RE = re.compile(
+    r'darwin-arm64.{0,50}darwin-x64.{0,120}win32-arm64.{0,50}win32-x64'
+)
 HOST_PLATFORM_THROW_RE = re.compile(
-    r'throw new Error\([^)]*Unsupported platform[^)]*\)'
+    r'throw new Error\([^)]*[Uu]nsupported.{0,20}platform[^)]*\)'
 )
 
 
 def patch_host_platform(filepath):
-    """Patch getHostPlatform() to return 'darwin-x64' instead of throwing on Linux.
-
-    The minified getHostPlatform() method only handles darwin and win32,
-    throwing Error('Unsupported platform: ...') for anything else.
-    Replace the throw with return"darwin-x64" so session init succeeds.
-    """
     with open(filepath, 'r') as f:
         content = f.read()
 
-    match = HOST_PLATFORM_THROW_RE.search(content)
-    if not match:
-        print(f"  getHostPlatform(): no throw found (already patched or not present)")
+    fp = HOST_PLATFORM_FINGERPRINT_RE.search(content)
+    if not fp:
+        print("  getHostPlatform(): fingerprint not found (build changed)")
         return True
 
-    content = HOST_PLATFORM_THROW_RE.sub('return"darwin-x64"', content)
+    region = content[fp.end():fp.end() + 500]
+    throw = HOST_PLATFORM_THROW_RE.search(region)
+    if not throw:
+        print("  getHostPlatform(): no throw near fingerprint (already returns darwin-x64)")
+        return True
+
+    abs_start = fp.end() + throw.start()
+    abs_end = fp.end() + throw.end()
+    content = content[:abs_start] + 'return"darwin-x64"' + content[abs_end:]
 
     with open(filepath, 'w') as f:
         f.write(content)
-
-    print(f"  getHostPlatform() patched: throw replaced with return\"darwin-x64\"")
+    print('  getHostPlatform() patched: throw -> return"darwin-x64"')
     return True
 
 
--- a/stubs/@ant/claude-swift/js/index.js
+++ b/stubs/@ant/claude-swift/js/index.js
@@ -1088,6 +1088,33 @@ class SwiftAddonStub extends EventEmitter {
       getEnabled: () => false,
     };
 
+    // Computer Use (TCC) — graceful denial on Linux. The asar accesses
+    // claude-swift.computerUse.* directly when its IPC fallback misses.
+    this.computerUse = {
+      getState: () => Promise.resolve({ accessibility: 'denied', screenCapture: 'denied', canPrompt: false }),
+      requestAccess: () => Promise.resolve({ success: false, accessibility: 'denied', screenCapture: 'denied', canPrompt: false }),
+      requestAccessibility: () => Promise.resolve({ success: false, accessibility: 'denied', canPrompt: false }),
+      requestScreenRecording: () => Promise.resolve({ success: false, screenCapture: 'denied', canPrompt: false }),
+      getCurrentSessionGrants: () => Promise.resolve([]),
+      revokeGrant: () => Promise.resolve(true),
+      openSystemSettings: () => Promise.resolve(null),
+      canPrompt: () => Promise.resolve(false),
+      listAvailablePermissions: () => Promise.resolve(['accessibility', 'screenCapture']),
+      listInstalledApps: () => Promise.resolve([]),
+      tcc: {
+        checkAccessibility: () => false,
+        checkScreenRecording: () => false,
+        requestAccessibility: () => false,
+        requestScreenRecording: () => false,
+      },
+    };
+
+    // Updater (ShipIt is macOS-only). Stub empty job info so the asar's
+    // updater logic short-circuits instead of throwing on null access.
+    this.updater = {
+      getShipItJobInfo: () => ({ label: null, isPrivileged: false }),
+    };
+
     // VM Management (nested object)
     // CRITICAL: The app accesses methods via module.default.vm, so all methods must be here
     const self = this;
@@ -1158,7 +1185,7 @@ class SwiftAddonStub extends EventEmitter {
           return { success: true };
         } catch {}
 
-        const arch = os.arch() === 'arm64' ? 'linux-arm64' : 'linux-x64';
+        const arch = process.arch === 'arm64' ? 'darwin-arm64' : 'darwin-x64';
         const url = 'https://downloads.claude.ai/claude-code-releases/' + encodeURIComponent(version) + '/' + arch + '/claude.zst';
         console.log('[claude-swift] Downloading SDK binary from ' + url);
         trace('vm.installSdk() downloading ' + url + ' -> ' + targetBin);
@@ -1251,6 +1278,17 @@ class SwiftAddonStub extends EventEmitter {
         );
 
         return spawnResult;
+      },
+
+      // Guest request interface — required by cliPluginBridge to classify CLI
+      // plugins. Without these the bridge skips init and dispatch reports
+      // "can't connect". (PR #112)
+      setGuestRequestCallback: (callback) => {
+        trace('vm.setGuestRequestCallback() registered');
+        self._guestRequestCallback = callback;
+      },
+      sendGuestResponse: async (id, resultJson, error) => {
+        trace('vm.sendGuestResponse() id=' + id);
       },
 
       kill: (id, signal) => {
--- a/stubs/@ant/claude-swift/js/index.js
+++ b/stubs/@ant/claude-swift/js/index.js
@@ -1088,8 +1088,8 @@ class SwiftAddonStub extends EventEmitter {
       getEnabled: () => false,
     };
 
-    // Computer Use (TCC) — graceful denial on Linux. The asar accesses
-    // claude-swift.computerUse.* directly when its IPC fallback misses.
+    // Computer Use — full namespace shape the asar accesses via hE().
+    // Linux has no Apple TCC; everything degrades to denied / empty.
     this.computerUse = {
       getState: () => Promise.resolve({ accessibility: 'denied', screenCapture: 'denied', canPrompt: false }),
       requestAccess: () => Promise.resolve({ success: false, accessibility: 'denied', screenCapture: 'denied', canPrompt: false }),
@@ -1107,6 +1107,29 @@ class SwiftAddonStub extends EventEmitter {
         requestAccessibility: () => false,
         requestScreenRecording: () => false,
       },
+      apps: {
+        listInstalled: () => Promise.resolve([]),
+        listRunning: () => Promise.resolve([]),
+        appUnderPoint: () => Promise.resolve(null),
+        iconDataUrl: () => undefined,
+        open: () => Promise.resolve(false),
+        unhide: () => Promise.resolve(),
+        prepareDisplay: () => Promise.resolve({ activated: null, hidden: [] }),
+        previewHideSet: () => Promise.resolve([]),
+        findWindowDisplays: () => Promise.resolve([]),
+      },
+      display: {
+        getSize: () => Promise.resolve({ width: 0, height: 0 }),
+        listAll: () => [],
+      },
+      // Mouse/keyboard primitives used by the executor — no-op on Linux.
+      key: () => Promise.resolve(),
+      moveMouse: () => Promise.resolve(),
+      mouseLocation: () => Promise.resolve({ x: 0, y: 0 }),
+      click: () => Promise.resolve(),
+      type: () => Promise.resolve(),
+      scroll: () => Promise.resolve(),
+      screenshot: () => Promise.resolve(null),
     };
 
     // Updater (ShipIt is macOS-only). Stub empty job info so the asar's
--- a/stubs/cowork/ipc_overrides.js
+++ b/stubs/cowork/ipc_overrides.js
@@ -386,67 +386,12 @@ function createOverrideRegistry(getProcessState) {
       return null;
     },
 
-    // ================================================================
-    // LocalAgentModeSessions — Dispatch/Bridge handlers
-    // ================================================================
-
-    'LocalAgentModeSessions_$_abandonBridgeEnvironment': async (_event, ...args) => {
-      vlog('[ipc:abandonBridgeEnvironment] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_deleteBridgeAgentMemory': async (_event, ...args) => {
-      vlog('[ipc:deleteBridgeAgentMemory] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_deleteBridgeSession': async (_event, ...args) => {
-      vlog('[ipc:deleteBridgeSession] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_getBridgeConsent': async (_event, ...args) => {
-      vlog('[ipc:getBridgeConsent] called');
-      return { consented: false };
-    },
-
-    'LocalAgentModeSessions_$_getSessionsBridgeEnabled': async () => {
-      return false;
-    },
-
-    'LocalAgentModeSessions_$_kickBridgePoll': async (_event, ...args) => {
-      vlog('[ipc:kickBridgePoll] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_onBridgePermissionPreflight': async (_event, ...args) => {
-      vlog('[ipc:onBridgePermissionPreflight] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_resetBridge': async (_event, ...args) => {
-      vlog('[ipc:resetBridge] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_resetBridgeSession': async (_event, ...args) => {
-      vlog('[ipc:resetBridgeSession] called');
-      return null;
-    },
-
-    'LocalAgentModeSessions_$_respondBridgePermissionPreflight': async (_event, ...args) => {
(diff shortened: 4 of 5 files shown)


michael-greider:
Oh shit I can't believe I forgot to mention it, the issue where the UI buttons for the side bar and such having a microscopic hit box is gone. That is probably the most revolutionary fix of all the Claude desktop Linux ports in existence

igor-raits:
I don't know if this is introduced in this PR or broken somewhere else (as the master version is not really working for me)… But I hit this issue with scheduled task:

```
==> /home/brain/.local/state/claude-cowork/logs/epipe-suppressed.log <==
[2026-05-09T09:51:05.739Z] EPIPE suppressed: Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    at writeGeneric (node:internal/stream_base_commons:150:3)
    at Socket._writeGeneric (node:net:1025:11)
    at Socket._write (node:net:1037:8)
    at writeOrBuffer (node:internal/streams/writable:570:12)
    at _write (node:internal/streams/writable:499:10)
    at Writable.write (node:internal/streams/writable:508:10)
    at CDi.write (/home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:397:9166)
    at /home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:399:5856
    at new Promise (<anonymous>)

==> /home/brain/.local/state/claude-cowork/logs/startup.log <==
[MAIN_LOG] 2026-05-09 11:51:05 [info] LocalAgentModeSessions.getSessionsForScheduledTask: scheduledTaskId=daily-summary
[MAIN_LOG] 2026-05-09 11:51:05 [warn] [HostLoop] cli.js stderr: /home/brain/.config/Claude/claude-code-vm/2.1.128/claude: /home/brain/.config/Claude/claude-code-vm/2.1.128/claude: cannot execute binary file
[MAIN_LOG] 2026-05-09 11:51:05 [error] Session local_45895207-ab77-4f43-ae27-1782b02a4c01 query error: Claude Code process exited with code 126 { stack: 'Error: Claude Code process exited with code 126\n    at CDi.getProcessExitError (/home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:397:8374)\n    at ChildProcess.i (/home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:397:11418)\n    at Object.onceWrapper (node:events:631:26)\n    at ChildProcess.emit (node:events:521:24)\n    at ChildProcess._handle.onexit (node:internal/child_process:295:12)' }
[claude-swift] vm.isGuestConnected() called - returning true
```

michael-greider:
@igor-raits thank you ! In short, there were two options in order to fix Anthropic blocking the OS check returning Linux-x64. Let the OS check return null and Electron natively has solid path resolution- which has performance issues and leaves gaps. Or return Darwin-x64 and manually solve all path resolution. ANY small issue like this that u see leave it in the comments or pm me and I'll incorporate it. Also, the reason manual redirection of Mac OS calls is better is having the full feature set actually working, it's basically trial and error until we hit full coverage. With every fix I make, I find new things that did not show up at all before. (These next are Not in this PR since I don't wanna bomb the repo with 20 commits, I'm gonna be comprehensively testing in arch hyprland, arch cage, and arch KDE plasma.) -Chrome connector works properly even with forks if chrome, helium in my case -live artifacts work properly -that annoying bs where UI buttons have a 1 pixel hit box -one major one is in the thinking lines of the chat, clause actually will navigate broken features and make it work at the cost of tokens

michael-greider:
@igor-raits additionally for your specific issue - Claude cowork uses its own Claude cli binary for tasks "frame-fix: route disclaimer to cowork SDK binary, not host claude" < commit #11 should fix it. If you would like (I recommend taking a snapshot or at least copying the full Claude cowork folder, I haven't had chats lost even once but better safe than sorry) you could clone my fork and just start letting it rip -- start from terminal and save the logs. Different use cases will reveal different issues, I for one use it as my code analysis agent since I do not like trusting AI fully with code. But if for example you use a lot of the JSX web app artifacts and such it would show us even more good info

igor-raits:
@michael-greider I am actually on your fork, I wiped whole Claude folder and when sending anything in cowork after login...

```
[MAIN_LOG] 2026-05-09 20:20:00 [info] [Lifecycle] Session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578: initializing → running
[MAIN_LOG] 2026-05-09 20:20:00 [info] [FileWatching] Starting file watcher for session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578: /home/brain/.config/Claude/local-agent-mode-sessions/f09ef3c0-0701-45a5-a9ea-df698c51ccf8/589d4cc9-bd84-477d-95bb-9080e26c3142/local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578/outputs
[MAIN_LOG] 2026-05-09 20:20:00 [info] [FileSystemWatcher] Started watching /home/brain/.config/Claude/local-agent-mode-sessions/f09ef3c0-0701-45a5-a9ea-df698c51ccf8/589d4cc9-bd84-477d-95bb-9080e26c3142/local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578/outputs for session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578

==> /home/brain/.local/state/claude-cowork/logs/epipe-suppressed.log <==
[2026-05-09T18:20:00.292Z] EPIPE suppressed: Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    at writeGeneric (node:internal/stream_base_commons:150:3)
    at Socket._writeGeneric (node:net:1025:11)
    at Socket._write (node:net:1037:8)
    at writeOrBuffer (node:internal/streams/writable:570:12)
    at _write (node:internal/streams/writable:499:10)
    at Writable.write (node:internal/streams/writable:508:10)
    at CDi.write (/home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:397:9166)
    at /home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:399:5856
    at new Promise (<anonymous>)

==> /home/brain/.local/state/claude-cowork/logs/startup.log <==
[MAIN_LOG] 2026-05-09 20:20:00 [warn] [HostLoop] cli.js stderr: /home/brain/.config/Claude/claude-code-vm/2.1.128/claude: /home/brain/.config/Claude/claude-code-vm/2.1.128/claude: cannot execute binary file
[MAIN_LOG] 2026-05-09 20:20:00 [error] Session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578 query error: Claude Code process exited with code 126 { stack: 'Error: Claude Code process exited with code 126\n    at CDi.getProcessExitError (/home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:397:8374)\n    at ChildProcess.i (/home/brain/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:397:11418)\n    at Object.onceWrapper (node:events:631:26)\n    at ChildProcess.emit (node:events:521:24)\n    at ChildProcess._handle.onexit (node:internal/child_process:295:12)' }
[claude-swift] vm.isGuestConnected() called - returning true
[MAIN_LOG] 2026-05-09 20:20:00 [info] [Lifecycle] Session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578: running → idle
[MAIN_LOG] 2026-05-09 20:20:00 [info] [FileWatching] Stopping file watcher for session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578
[MAIN_LOG] 2026-05-09 20:20:00 [info] [FileSystemWatcher] Stopped watching for session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578
[MAIN_LOG] 2026-05-09 20:20:00 [debug] [QueryLoop] Exited for session local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578 after 5ms, exitPath=error: Claude Code process exited with code 126, lifecycleState=idle
[MAIN_LOG] 2026-05-09 20:20:00 [info] [CycleHealth] Unhealthy cycle: {
  session_id: 'local_94cb9f11-8fb7-47fe-9e68-9e7bdc702578',
  scheduled_task_id: null,
  space_id: undefined,
  session_type: undefined,
  parent_session_id: undefined,
  host_loop_mode: true,
  vm_instance_id: '1cdec0f5-14af-40d7-b906-ad3f556d9721',
  model: 'claude-opus-4-7[1m]',
  cli_session_id: null,
  user_message_uuid: '4c740683-841d-4d22-bab2-736d5923ce86',
  cycle_health: 'unhealthy',
  had_first_response: false,
  seconds_to_outcome: 14,
  is_resume: false,
  unhealthy_reason: 'system_error',
  error_message: 'Claude Code process exited with code 126',
  error_category: 'process_crashed',
  transcript_size_bytes: undefined
}
```

michael-greider:
@igor-raits working on that now, no need to wipe the cowork folder ! There's only 3 u need to clean between installs. There's gpucache , dawnwebgpu cache and .config/Claude/claude-code/2.1.128/claude.app -- what you're missing there since the deletion of the full folder is the claude-code-vm/2.1.128/ folder needs a claude binary separate from your system one. My fault playa should've said that, I didn't include the installation of it in my fork because with the path resolution was pointing towards the first it could find, which is your normal /opt/claude-code. I'll push the fixes shortly. Fr my bad though I'm autistic as shit and could not sit with the extremely minor errors that came from letting the fallback to the root level binary happen. Pushing some more fixes soon that will be segmented across 2 PRs, 1 is a complete option and allows for computer use to work and routes the currently non working System settings to a qtquick2 mini widget that configures the needed VNC and screen recording protocols, it also routes the folder access popup that controls what folders Claude cowork can access. Like if you're working in~/Projects but need it to take a look at ~/Downloads , natively as long as Downloads is an allowed folder it can look without needing that folder added to the project. At the moment it's either your whole OS or just the project folder. The other is locking in the functionalities like the isolated binary (what you're missing) and tool use permission requests mishandling. As well as properly isolating the binary as to not conflict with your Claude code install. It doesn't break anything, but it will burn tokens when cowork attempts a tool use incorrectly and has to try it again due to the gap from Claude code tool to cowork tool. PM me if u need a binary for Claude code, my big blocker is making the separation DMCA approved but there's another solution that is 100 percent NOT demincing the isolated claude code ELF binary and stripping path and setting resolutions🫣

michael-greider:
@igor-raits also look at #107 in issues, I listed all the path breakages from V5.0.0 -- I'm going for extremely ambitious coverage of everything that Mac OS expects. That's just because I highly suspect that unless there's full coverage things will always be silently breaking or causing data loss and tool use problems. So if you're really trying to get it working ASAP and not having issues, use the V5.00 install, then use my fork to update it, and get desktop commander MCP to cover tool use gaps. You could also just add the Claude code binary installation from 5.0.0 back in

johnzfitch (changes requested):
Big improvement on the meta-issue overall — `enable-cowork.py` fingerprinting, macOS-only command rerouting, disclaimer version-pinning, the `app.dock` namespace, and the arch un-spoof are all real wins, and subsuming #108/#103/#110/#111/#112 cleans the PR queue. A few items below; one I want to block on.

### Block — bridge IPC override removal (`stubs/cowork/ipc_overrides.js` lines 386-450)

Removing the 12 bridge handler stubs lets the asar's `LocalAgentModeSessionManager` open a persistent WebSocket to `wss://bridge.claudeusercontent.com`. That's a second persistent network egress beyond the existing `claude.ai`/`auth.anthropic.com` HTTP API, with its own protocol surface to audit and re-audit on each upstream update. The "one networking boundary" posture is intentional.

Proposing we gate the WebSocket itself rather than re-stub the IPC, so the rest of `LocalAgentModeSessionManager` initializes correctly. Add to `frame-fix-wrapper.js` after the SingleInstance block:

```js
// Block the dispatch WebSocket. The asar opens this when sessions-bridge
// init succeeds; we route dispatch through the existing API boundary
// instead. One networking boundary, not two.
try {
  const { app, session } = require('electron');
  app.whenReady().then(() => {
    const s = session.defaultSession;
    if (!s) return;
    s.webRequest.onBeforeRequest(
      { urls: ['*://bridge.claudeusercontent.com/*'] },
      (_details, callback) => callback({ cancel: true })
    );
    console.log('[bridge-gate] Installed');
  });
} catch (e) { console.error('[bridge-gate] Failed:', e.message); }
```

Open question to confirm before merge: does `tool_permission_request` have a local-only path (main↔renderer IPC) when the WebSocket can't connect? If yes, this gate is the cleanest fix and the existing `forwardEvent` patch covers result/stream_event without the bridge. If no, we should either pin the bridge URL in an explicit allowlist + document the protocol/data flow in the PR description, or build a Linux-native permission UX.

### Request changes (security-relevant, smaller)

1. **`_macOnlyCmdNoop` `open` reroute** (`frame-fix-wrapper.js` ~line 720) — no scheme validation. `xdg-open file:///etc/passwd` or `claude://...` would both forward. Restrict to http/https/mailto:

   ```js
   if (base === 'open') {
     const target = args.filter(a => !a.startsWith('-')).pop();
     if (!target) return { cmd: '/bin/true', rest: [] };
     try {
       const p = new URL(target);
       if (!['http:','https:','mailto:'].includes(p.protocol)) return { cmd: '/bin/true', rest: [] };
     } catch { return { cmd: '/bin/true', rest: [] }; }
     return { cmd: 'xdg-open', rest: [target] };
   }
   ```

2. **Disclaimer version-string path traversal** (`frame-fix-wrapper.js` ~line 683). `[^/]+` allows `..`; `path.join` happily traverses one level up. Pin to the expected prefix:

   ```js
   const vmRoot = path.join(xdgCfg, 'Claude', 'claude-code-vm');
   const vmBin = path.normalize(path.join(vmRoot, m[1], 'claude'));
   if (!vmBin.startsWith(vmRoot + path.sep)) return null;
   ```

3. **`askForMediaAccess` early stub** (`frame-fix-wrapper.js` ~line 56) returns `Promise.resolve(true)` but the existing `Module.prototype.require` interceptor reassigns it to `false` later. Last-write-wins, so it's dead code at runtime — but confusing, and a future move of the interceptor would silently flip the default to "user granted camera/mic". Set the early stub to `false` to match, or delete it.

### Questions

4. **SDK URL change to `darwin-x64`** (`stubs/@ant/claude-swift/js/index.js` ~line 1211). We now download a Mach-O that `launch.sh`'s symlink swap immediately overwrites. Was the linux URL deprecated, or is this matching the spoofed `getHostPlatform()`? If the latter, prefer `linux-x64` and drop the symlink swap.

5. **PKGBUILD cache wipe on every launch** (lines 251-260). Kills GPU/font caches that take real time to rebuild. Worth gating on a version marker (`stat $_CFG/.cache-version` vs `$pkgver`) so it fires only on asar version changes, not every cold start.

### Approve — rest of the diff

`enable-cowork.py` fingerprint anchor over the old global-sub is the right shape. `xcrun git → /usr/bin/git` special case is exactly right. `app.dock` namespace, `BrowserWindow.prototype.setHiddenInMissionControl`, and the activity-API stubs (subsumes #108) close the right holes. Layout mirror in PKGBUILD makes both resourcesPath lookups resolve naturally. Computer-use namespace stub fails-closed correctly — and your call to defer the actual Linux wiring is the right one.

---

michael-greider:
@johnzfitch on the money for the most part -- there's slight issues with the permission requests and I'm autistic as shit so I keep trying to make it perfect. I'm gonna get to my computer later tonight, I have the stable things written down separately. My fault playa, I got wayyyy too caught up with DMCA approved play dates with the binary.  

I've gone down 3 pathways of experimenting. Make it stable with minimal changes, make it stable with full coverage, make everything work exactly how it does on Mac because I get too hyper focused and am dumb

*overall things that I see on all 3* 

1. Settings not persisting, tool_permission_request has been weird to pin down, it's a matter of the model seeing tools it can't access. The forwardEvent pass for me at least was having permission requests just fall through- however you CAN get it to work properly, you just also need to setup the computer use permission access. I actually have a QML sidecar (more proposal level, still has kinks) that has the same UI look as the app. ydotool grim and slurp is all it needs. The underlying issue that came about with forwardEvent pass was the model thinking it had computer use, trying to use it, and causing performance issues in terms of tokens (the horrendous claude code database has made these models spectacular at working around broken tools and using them right, bad when it's non existent). < also for this one, I was trying to not be a dick to everybody else- since we're both (arch hyprland btw) on the same daily driver (which is arch btw) and work out computer use to the direct Wayland IPC bindings. That can make all permission requests (I have no idea how, this was a close the laptop for the night level of annoyance for me) route through GTK popups, I'm assuming that's electron doing its thing but basically it can cause broken tool use recursion that burns some tokens or we pin down where the differentiation between actual settings that involve permissions and the requested permissions in chats. The main issue is tool use attempts for things that aren't there 

There is also of course, making computer use work as it does on Mac, pixel buffer stream and cursor events, the two Mac system settings in chat and in the actual settings are cursor access and screen recording (which is just the buffer streaming) + mic access is the accessibility setting when you press [Open System Settings] < not needed for cu but works beautifully 

^this is the main thing that hung me up, I kept finding that the more native features i dug out and made work the more stable shit gets. (I'll edit screenshots onto this post when I'm home) 


2. There's some kind of fingerprinting mechanism with the way projects are made that I kept having popup, basically just the entry with the name of the project disappears - all of its actual associated tool use data, caching, and chat data stays but very annoying and worth noting 

3. Electron is a much bigger nuisance than I initially thought it would be. A lot of systematic issues that stem from its native path resolution + having to also deal with anthropics "lets ruin the day of everybody who runs Linux" related issues. Command invocations like bash for example, easily get broken with even the smallest changes. File system navigation, forwardEvent pass specifically made this very odd issue of claude accessing a directory outside of the current one (on request) > it looks at that directory > working directory gets thrown to the config dir. 

*when I get home I will pull up what I've found, if you're in the ide right now tryna get shit done definitely just try the forwardEvent pass first, then when you start working out minor issues you'll understand what I'm getting at with a lot of the issues being uncanny and hard to pin down. My goal kinda had shifted to full send getting every path resolution and IPC call working in some way- because in my experience that is the easiest position to then find where the things natively resolved by electron and the things expecting Mac calls nutkick each other*

michael-greider:
^^all in all my fault playa, the only thing that definitely gets fucky is the forwardevent pass. But I tismd too hard and overscoped -- will send forward some notes and testing asap, was trying to wrap things up and push absolute fixes before it got reviewed to be honest- should have marked this PR for testing only

johnzfitch:
Sounds good. It’s all convoluted and tricky, I know how difficult it is to work with electron. Anthropic doesn’t make it any easier by shipping layer upon layer of spaghetti code.

michael-greider:
@johnzfitch I'm gonna have to get back to you on this one, Speaking of anthropic not making things easier i had it running a long list of tests while i was at work. All just trying thing's to map out the ABI interactions and platform calls. It took the very ambitious initiative to skip out on the 108 test gates, and was even courteous enough to traverse from my .cache where i had the copy test repo and wiped my real one. There was some issues with it's ability to merge directly to this repo, so it kindly rebased with 497 commits and was even security conscious enough to delete the documentation before doing it! i'm gonna work on full DMCA friendly play date of the binary - i've already done it with the new ELF for claude code. I am so done with Anthropic's shit. i'll send over the reconstructed research - can't trust their models to run 4 scripts and read at the moment.

johnzfitch:
Hey -- v5.1.0 merged (#108) and took the bridge override removal + activity stubs that overlapped with this PR. The parts of #113 that are still unique and worth landing:

- `enable-cowork.py` fingerprint-based getHostPlatform patch (better than the global-sub)
- `claude-swift/index.js` computerUse namespace + updater.getShipItJobInfo + guest request methods
- macOS-only command rerouting (security/xcrun/osascript/open)
- arch un-spoof (process.arch returns real value)
- disclaimer-arg version-pinning for mcp__workspace__bash

The three items from my review still apply to what's left: `/usr/bin/open` scheme validation, disclaimer version-string path normalization, `askForMediaAccess` early stub returning true instead of false.

If you rebase onto current master, the bridge-handler diff drops out and the remainder is a clean feature-coverage PR. No rush -- just noting the state so you know what's already in.

---

PR #115, opened 2026-05-11
fix: stub macOS NSUserActivity app methods to prevent crash on Linux

bpainter64:
## Problem

When the user selects certain UI options (e.g. navigating between tabs/views), the asar calls `app.invalidateCurrentActivity()` — part of Electron's macOS-only [NSUserActivity API](https://www.electronjs.org/docs/latest/api/app#appinvalidatecurrentactivity-macos). On Linux this method does not exist, producing an uncaught `TypeError` that crashes the main process:

```
Uncaught Exception:
TypeError: gA.app.invalidateCurrentActivity is not a function
    at Frt (.../app.asar/.vite/build/index.js:2467:9879)
    at $ir (.../app.asar/.vite/build/index.js:2467:10193)
    at WebContents.t (.../app.asar/.vite/build/index.js:3157:24129)
    at WebContents.emit (node:events:521:24)
```

The app spoofs `process.platform = 'darwin'`, so the asar's platform guard doesn't protect these calls.

## Fix

Stub the four NSUserActivity `app` methods as no-ops in `frame-fix-wrapper.js`:

- `invalidateCurrentActivity`
- `updateCurrentActivity`
- `setCurrentActivity`
- `getCurrentActivityType`

**Two insertion points** are used, mirroring the existing `systemPreferences` stub pattern:

1. **Early patch** (top of file, before `Module.prototype.require` override) — destructures `app` alongside `systemPreferences` from the initial `require('electron')` call. This ensures stubs are in place before any asar code runs.

2. **Late patch** (inside the `require('electron')` intercept block) — belt-and-suspenders for any code that requires electron after the override is installed.

## Testing

- Launch app → select options that trigger navigation/tab switches → no crash dialog
- Verified on Linux with Electron 33 / app.asar from Claude Desktop

### DIFF
--- a/stubs/frame-fix/frame-fix-wrapper.js
+++ b/stubs/frame-fix/frame-fix-wrapper.js
@@ -19,7 +19,7 @@
 }
 
 // Patch macOS-only systemPreferences methods that don't exist on Linux
-const { systemPreferences } = require('electron');
+const { systemPreferences, app: _earlyApp } = require('electron');
 if (typeof systemPreferences.setUserDefault !== 'function') {
   systemPreferences.setUserDefault = function(key, type, value) {
     // no-op on Linux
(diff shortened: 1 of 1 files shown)


johnzfitch:
Overlapped with #108 which landed today, same stubs there plus a few extras. Thanks for the careful PR.

---

PR #117, opened 2026-05-16
fix: x86_64 SDK download and macOS Handoff API crash

Medstaff-excel:
## Summary

Two bugs that prevent Cowork from working on x86_64 Linux:

- **Wrong SDK architecture**: `installSdk()` in the swift stub uses `os.arch()` to detect platform, but `frame-fix-wrapper.js` spoofs `process.arch` to `'arm64'` for macOS compatibility. This causes the ARM64 binary to be downloaded on x86_64 machines. Fixed by using `uname -m` to detect real architecture.

- **Handoff API crash**: `app.invalidateCurrentActivity()` and `app.setUserActivity()` are macOS-only Electron APIs (NSUserActivity/Handoff). The minified bundle calls these unconditionally when `process.platform === 'darwin'`, which is always true due to platform spoofing. On Linux these are undefined, causing a `TypeError` crash dialog on every launch. Fixed with no-op stubs in two layers: sed patch in `launch.sh` and runtime stub in `frame-fix-wrapper.js`.

## Test plan

- [ ] Install on x86_64 Linux, verify `~/.config/Claude/claude-code-vm/*/claude` is an x86-64 ELF binary (not ARM64)
- [ ] Launch claude-desktop, verify no "invalidateCurrentActivity is not a function" error dialog
- [ ] Send a message in Cowork tab, verify bash oneshot spawns succeed (exit code 0)
- [ ] Verify on ARM64 Linux that `uname -m` returns `aarch64` and correct binary is downloaded

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -115,6 +115,14 @@ if [ -f "$INDEX_JS" ] && grep -q 'e\.protocol==="file:"&&Ee\.app\.isPackaged===!
   sed -i 's/e\.protocol==="file:"&&Ee\.app\.isPackaged===!0/e.protocol==="file:"/g' "$INDEX_JS"
 fi
 
+# Fix macOS Handoff API: invalidateCurrentActivity() and setUserActivity() are
+# macOS-only Electron APIs that crash on Linux. Replace with safe no-op fallbacks.
+if [ -f "$INDEX_JS" ] && grep -q 'cA\.app\.invalidateCurrentActivity()' "$INDEX_JS"; then
+  echo "Patching macOS Handoff APIs for Linux..."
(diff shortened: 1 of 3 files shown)


johnzfitch:
Handoff API stubs are covered by v5.1.0 (commit f93a0be in #108). The x86_64 SDK download arch bug is still real though -- `os.arch()` returns `'arm64'` due to spoofing. That half is worth extracting as a standalone fix or rebasing onto current master.

---

johnzfitch:
Closing this since the Handoff stubs are in v5.1.0. The SDK arch fix (os.arch() returning arm64 due to spoofing) is still a real bug -- if you want to extract just that piece into a fresh PR against current master, it'd be a clean single-file diff in claude-swift/index.js. The fix shape (use uname -m or process.arch after the arch un-spoof) is right.

---

PR #121, opened 2026-05-17
fix(cowork): implement CoworkSpaces persistence on Linux

Medstaff-excel:
## Summary

- **Spaces/Projects now persist across restarts** on Linux
- Root cause: the asar's native SpacesStore registers IPC handlers only after account details arrive from the renderer, which is unreliable on Linux — without handlers, `ipcRenderer.invoke('getAllSpaces')` fails silently
- Fix: register all 21 CoworkSpaces IPC handlers on `ipcMain.handle()` at startup before any webContents exists, backed by a file-based `spaces_store.js` that reads/writes `spaces.json`
- When the native handler eventually registers on `webContents.ipc` (if account init succeeds), it takes priority automatically

## Key insight

Electron's `webContents.ipc.handle()` and `ipcMain.handle()` are **independent dispatch tables** — `ipcMain` is NOT a fallback when `webContents.ipc` has no handler. The fix registers early on `ipcMain` with the known EIPC UUID so there's always a handler available.

## Changes

| File | Change |
|------|--------|
| `stubs/cowork/spaces_store.js` (new) | File-backed CRUD for all 21 CoworkSpaces IPC methods, persisting to `spaces.json` |
| `stubs/frame-fix/frame-fix-wrapper.js` | Early `ipcMain.handle()` registration with hardcoded EIPC UUID |
| `stubs/cowork/ipc_overrides.js` | CoworkSpaces override registry + proactive registration list |
| `stubs/cowork/session_orchestrator.js` | Recover `sharedCwdPath` from session metadata when asar passes `false` |
| `launch.sh` | `mainView.js` origin patch for `file://` protocol (idempotent) |

## Test plan

- [x] Launch Cowork, create a project (pin a folder) — project appears in sidebar
- [x] Close Cowork, relaunch from app launcher — project persists with context/instructions
- [x] Projects page shows all spaces
- [x] Instructions and folder context survive restarts
- [x] No duplicate spaces created on restart

### DIFF
--- a/launch.sh
+++ b/launch.sh
@@ -115,6 +115,18 @@ if [ -f "$INDEX_JS" ] && grep -q 'e\.protocol==="file:"&&Ee\.app\.isPackaged===!
   sed -i 's/e\.protocol==="file:"&&Ee\.app\.isPackaged===!0/e.protocol==="file:"/g' "$INDEX_JS"
 fi
 
+# Fix preload origin validation: the mainView.js preload's h() guard checks
+# if window.location.href origin matches claude.ai/preview.claude.ai etc.
+# On Linux with file:// protocol, origin is "null" and h() returns false,
+# preventing CoworkSpaces (and other IPC bridges) from being exposed to the
+# renderer. This causes the Projects page to be empty and spaces to not persist.
+# Patch: add file:// protocol as an allowed origin.
+MAINVIEW_JS="linux-app-extracted/.vite/build/mainView.js"
+if [ -f "$MAINVIEW_JS" ] && ! grep -q 'e\.protocol==="file:"' "$MAINVIEW_JS"; then
+  echo "Patching preload origin validation for file:// protocol..."
+  sed -i 's/e\.hostname==="localhost"/e.hostname==="localhost"||e.protocol==="file:"/g' "$MAINVIEW_JS"
+fi
+
 # Fix resource path lookup for i18n, shim-lib, icon, etc.
 # The asar uses `app.isPackaged ? process.resourcesPath : <asar-relative path>`.
 # On Arch Linux, `process.resourcesPath` is the system electron's dir
--- a/stubs/cowork/ipc_overrides.js
+++ b/stubs/cowork/ipc_overrides.js
@@ -20,6 +20,7 @@ const {
   COMPUTER_USE_TCC_GRANTED,
   COMPUTER_USE_TCC_REQUEST_GRANTED,
 } = require('./linux_ipc_stubs.js');
+const { createSpacesStore } = require('./spaces_store.js');
 
 // -- Helpers --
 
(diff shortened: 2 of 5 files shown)


johnzfitch (changes requested):
Good feature — spaces persistence is real and #123 confirms users need it. A few things to address:

**Security (request changes):**

1. `readFileContents(_event, filePath)` reads ANY file with no path validation. An IPC call from the renderer could read `/etc/shadow`, `~/.ssh/id_rsa`, etc. Need to validate that `filePath` is within one of the space's registered folders or within localAgentRoot.

2. `listFolderContents(_event, folderPath)` same issue — lists any directory. Validate against allowed roots.

3. `openFile(_event, filePath)` calls `xdg-open` on any path. Should at minimum validate the path is within a space folder, and ideally restrict to known safe file types.

The pattern from the existing code is `isPathWithinAllowedRoots()` in ipc_overrides.js — reuse that, or at least check `path.resolve(filePath).startsWith(homedir)`.

**Fragility:**

4. The hardcoded EIPC UUID (`c42bb833-f6f9-4563-9861-03480449cfb1`) in `frame-fix-wrapper.js` line 505 will break silently on the next asar build. The existing code discovers UUIDs dynamically via `extractEipcUuid()` — use the same pattern. The proactive registration in ipc_overrides.js already handles this (it's in `PROACTIVE_ONLY_SUFFIXES` which gets registered when the first EIPC channel is seen).

Since you already added the 21 methods to `PROACTIVE_ONLY_SUFFIXES`, the proactive registration in `frame-fix-wrapper.js`'s `ipcMain.handle` interceptor should pick them up automatically when the asar registers its first EIPC channel. The hardcoded early block at line 479-547 is redundant and can be removed — test whether spaces persist without it.

**Minor:**

5. The launch.sh patch for `mainView.js` origin validation overlaps with #119 (IPC origin bypass). If #119 lands first, this sed might conflict or be redundant. Worth checking.

6. `session_orchestrator.js` sharedCwdPath recovery is a good fix and independent of the spaces work. Consider splitting it out as a separate small PR so it can land without the security discussion blocking it.

The spaces_store.js module itself is clean — good factory pattern, proper JSON round-tripping, dedup on addFolder, id/createdAt protection on update. Just needs the path validation gate on the three filesystem-access methods.

---

johnzfitch:
Closing as superseded by #127, now merged to `master`. #127 explicitly consolidates this PR — CoworkSpaces persistence landed there as `stubs/cowork/spaces_store.js` (file-backed CRUD with 21 IPC handlers), the early `ipcMain.handle()` fallback registration, and the `mainView.js` `file://` origin patch.

Thanks for the work and the IPC dispatch-table insight — it shipped via #127. If anything from here didn't make it in, please reopen or file a follow-up.

---

Issue #128, opened 2026-05-24
Crash on launch: gA.app.configureWebAuthn is not a function (asar 1.8555.2)


brucebruce603:
### Area

Install / setup

### What happened?

Fresh install completes successfully, but the app crashes during main-process load before any window appears — no GUI ever opens. Expected a Claude Desktop window; instead the shell just returns after two [WARN] version lines.

app.configureWebAuthn() exists in Electron 42, so this isn't a missing-API problem on my end. The 1.8555.2 main process calls it in a way that throws under the current frame-fix wrapper. Cowork patch and Swift stub both report applied; --doctor passes 16 checks (only failures: /sessions symlink, blocked by read-only ostree root, and the expected newer-than-tested asar warning).

Related to #114 (also newer-asar breakage) but fails earlier — at launch, not in /setup-cowork — and on a newer asar.

### Steps to reproduce

1. git clone + ./install.sh --force on Bazzite (Electron 42.2.0)
2. Installer auto-fetches asar 1.8555.2 from CDN
3. Run claude-desktop
4. Two [WARN] version lines print, shell returns, no window appears
5. startup.log shows the configureWebAuthn TypeError

### Relevant log output

```shell
[Platform] Spoofing: darwin/arm64 macOS 14.0 (immediate)
[Cowork] Linux support enabled - VM will be emulated
Starting app { appVersion: '1.8555.2', isPackaged: false,
  platform: 'darwin', arch: 'arm64', nodeVersion: '24.15.0' }
[claude-native-stub] claude-native stub loaded successfully
[Frame Fix] Stubbed systemPreferences.setUserDefault: NSAutoFillHeuristicsEnabled
App threw an error during load
TypeError: gA.app.configureWebAuthn is not a function
    at uJr (/home/USER/.local/share/claude-desktop/.asar-cache/app.asar/.vite/build/index.js:3337:36189)
    at Object.<anonymous> (.../app.asar/.vite/build/index.js:3337:38503)
    at Module._compile (node:internal/modules/cjs/loader:1838:14)
    at Module.require (.../app.asar/frame-fix-wrapper.js:1018:34)
A JavaScript error occurred in the main process
Uncaught Exception:
TypeError: gA.app.configureWebAuthn is not a function
```

### Distro and desktop environment

Bazzite (Fedora Atomic 43, immutable/ostree) / KDE Plasma (Wayland)

### claude-cowork-linux version / commit

[...]

PR #129, opened 2026-05-25
fix: stub configureWebAuthn no-op for asar 1.8555.2

ndaube:
## What
Stubs `app.configureWebAuthn()` as a no-op so asar 1.8555.2 starts on Linux.

## Why
asar 1.8555.2 calls `app.configureWebAuthn()` early in startup. Linux Electron 
doesn't have this method, so the main process crashes before any window appears.

## Tested
- Ubuntu 24.04, GNOME/Wayland
- asar 1.8555.2
- App opens and Cowork tab is functional

- [ ] This change does touch security-sensitive code — explanation below:

## Checklist

- [ ] No API keys, tokens, `.env` files, or log output with credentials included
- [ ] Commit messages follow the project style (no emoji, brief summary + explanation)
- [ ] `./install.sh --doctor` passes cleanly on my system

### DIFF
--- a/stubs/@ant/claude-native/index.js
+++ b/stubs/@ant/claude-native/index.js
@@ -314,6 +314,8 @@ module.exports = {
   get_app_info_for_file,
   getAppInfoForFile: get_app_info_for_file,
 
+  // WebAuthn (no-op on Linux)
+  configureWebAuthn: () => {},
   // Native stub functions
   ...nativeStub,
 };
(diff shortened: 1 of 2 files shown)


johnzfitch:
Closing as superseded. The `configureWebAuthn` crash (#128) is now fixed on `master` via #131, which adds `'configureWebAuthn'` to the existing `_macOnlyAppMethods` no-op loop in `frame-fix-wrapper.js` — the same place the Handoff/NSUserActivity stubs live.

This PR's approach was correct; the maintainer opted for the single-line addition to the existing guarded loop rather than the two-file version (the crash is on the Electron `app` object, which that loop already handles, so the `@ant/claude-native` export edit wasn't needed). Thanks for diagnosing it against asar 1.8555.2 — that's the exact version that confirmed the fix. If you spot a gap on current `master`, please reopen.

---

PR #130, opened 2026-05-30
stub app.configureWebAuthn for Linux (asar 1.9659.2+)

kenjoe41:
## Problem

asar v1.9659.2 calls  during main-process initialisation. This is a macOS-only Electron API — on Linux Electron it does not exist, causing an uncaught  that crashes the app before any window opens:

## Fix

Add  to the  no-op list in , alongside the existing NSUserActivity / Handoff stubs (issues #104, #106). The method is stubbed to a no-op so the app continues without WebAuthn hardware-key support (not relevant on Linux).

## Testing

- Reproduced crash locally with asar 1.9659.2 + Electron v42.3.0 on Wayland (GNOME)
- Applied fix → app launches cleanly, all cowork rails engaged
- No regression on existing features (chat, Code tab, cowork sessions)

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -5,13 +5,14 @@ to work with claude-cowork-linux. `install.sh` and the launcher grep the
 machine-readable lines below; the table further down is for humans.
 
 <!-- machine-readable; do not remove the next two lines -->
-<!-- LAST_TESTED_ASAR_VERSION=1.6259.1 -->
-<!-- LAST_TESTED_DATE=2026-05-14 -->
+<!-- LAST_TESTED_ASAR_VERSION=1.9659.2 -->
+<!-- LAST_TESTED_DATE=2026-05-30 -->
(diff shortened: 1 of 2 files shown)


johnzfitch:
Closing as superseded. The `configureWebAuthn` crash (#128) is now fixed on `master` via #131, which adds `'configureWebAuthn'` to the existing `_macOnlyAppMethods` no-op loop in `frame-fix-wrapper.js`.

This PR carried the same wrapper edit plus a `COMPAT.md` bump to 1.9659.2. The code fix landed via #131; the COMPAT bump was held back since 1.9659.2 was marked `[PARTIAL]` (launch smoke-test only, full feature exercise pending) — worth a separate PR once login/sessions/MCP have been verified on that version. Thanks for testing against the newer asar. If the crash persists on current `master`, please reopen.

---

PR #141, opened 2026-06-15
fix(swift): vm.spawn() returns a promise resolving to { failedMounts }

davide-perego-runtime:
## Problem

On asar v1.5xx+ the app awaits `vm.spawn()`'s return value and calls `.catch()` on it, then destructures `{ failedMounts }`:

```js
const o = vm.spawn(...); o.catch(()=>{});
await Promise.race([o, waitForExit()]);
const { failedMounts = [] } = await o;
```

The stub returned a plain object, so every workspace bash command (resume/create/re-resume) crashed with `o.catch is not a function`.

## Fix

- `vm.spawn()` now returns a Promise.
- Rejects on prepare/spawn failure (with `failedMounts: []` attached) so the error surfaces immediately instead of hanging until the 30s oneshot timeout.
- Otherwise resolves to `{ ...spawnResult, failedMounts: [] }` — the shape the asar destructures.
- Keeps `success`/`pid`/`failedMounts` attached synchronously on the returned promise so existing non-awaiting callers and the node test harness keep working.

### DIFF
--- a/stubs/@ant/claude-swift/js/index.js
+++ b/stubs/@ant/claude-swift/js/index.js
@@ -1294,8 +1294,21 @@ class SwiftAddonStub extends EventEmitter {
           sharedCwdPath,
           onError: self._onError,
         });
+        // The asar (v1.5xx+) treats vm.spawn()'s return value as a Promise:
+        //   const o = vm.spawn(...); o.catch(()=>{});
+        //   await Promise.race([o, waitForExit()]);
+        //   const { failedMounts = [] } = await o;
+        // Returning a plain object here makes the app crash with
+        //   "o.catch is not a function"
+        // on every workspace bash command (resume/create/re-resume all fail).
+        // So spawn must resolve to an object exposing `failedMounts`, and
+        // reject when the process cannot be launched so the error surfaces
+        // instead of hanging until the 30s oneshot timeout.
         if (!preparedSpawn.success) {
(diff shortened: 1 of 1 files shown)


johnzfitch (approved):
Verified locally and the fix is correct.

**Root cause confirmed:** current `vm.spawn()` returns a plain object; the asar v1.5xx+ does `o.catch(()=>{})` / `await o` / `const { failedMounts = [] } = await o`, so every workspace bash command (resume/create/re-resume) crashes with `o.catch is not a function`.

**Testing (no CI runs these, so I ran them by hand):**
- Full node suite — **493/493 pass** (`node --test tests/node/current-path/*.test.cjs`), including the spawn-sensitive suites: `claude_swift_resume_retry` (reads `spawnResult.success` synchronously) and `security_hardening` (calls `vm.spawn` and ignores the now-Promise return).
- Focused contract check mirroring the asar's exact consumption against a real spawn: returned value is a thenable (`.then`/`.catch`), `await o` yields `{ failedMounts: [] }`, `success`/`pid` are exposed synchronously, and the prepare/spawn-failure path returns a *rejecting* thenable. All green.

The hybrid promise (resolved value for `await`, plus synchronous `success`/`pid`/`failedMounts`) is the right bridge between the new asar contract and the existing synchronous callers/tests. Merging.

---

PR #145, opened 2026-06-20
fix(spaces): match createSpaceFolder to new (parentPath, folderName) contract

### LINKED ISSUE #144: Project creation fails: createSpaceFolder rejects new (parentPath, folderName) contract
## Symptom

Every attempt to create a project fails with a toast: **"Failed to create project. You can try again."**

`~/.local/state/claude-cowork/logs/startup.log` shows:

```
[spaces] createSpaceFolder BLOCKED (invalid spaceId): /home/<user>/Claude/Projects
```

## Root cause

Asar version drift. Installed asar is **1.14271.0**; the port was last tested against **1.6259.1** (`COMPAT.md`).

The CoworkSpaces stub (`stubs/cowork/spaces_store.js`) implements the **old** `createSpaceFolder` contract:

```js
createSpaceFolder(_event, spaceId, folderName)  // arg0 must be a UUID
// → creates localAgentRoot/spaces/<uuid>/<folderName>
```

But in 1.14271.0 the "New Project" flow calls it with a **filesystem path** as arg0. The bundle's own IPC validator confirms the new shape:

```
Argument "parentPath" at position 0 to method "createSpaceFolder" ...
Argument "folderName" at position 1 to method "createSpaceFolder" ...
```

…and the native impl is effectively:

```js
async createSpaceFolder(parentPath, folderName) {
  let t = join(parentPath, folderName.trim());
  // collision dedup: "name (1)", "name (2)", …
  await mkdir(t, { recursive: true });
  return t;
}
```

So the absolute path `~/Claude/Projects` lands in the stub's UUID check, fails `isValidUUID`, and the call is rejected — project creation never happens.

## Fix

PR #145 — rewrites the stub to the new `(parentPath, folderName)` contract: creates `<parentPath>/<folderName>`, dedups on collision to match the native impl, returns the created path. The native relies on the macOS sandbox for confinement; Linux has none, so creation is confined to the allowed roots (homedir, no `/tmp`) via the existing `requireAllowedPath` (same policy as `addFolderToSpace`), validating the normalized target so a `..`-laden parent can't escape.

This is a COMPAT regression from the version jump, independent of the MCP node-host fix (#142 / #143).

shawnyeager:
## Problem

Creating a project fails with "Failed to create project." Log:

```
[spaces] createSpaceFolder BLOCKED (invalid spaceId): /home/<user>/Claude/Projects
```

Fixes #144.

## Root cause

The CoworkSpaces stub implements the old `createSpaceFolder(spaceId, folderName)` contract — arg0 validated as a UUID, folder created under `localAgentRoot/spaces/<uuid>/<name>`. In asar **1.14271.0** the "New Project" flow calls it as `createSpaceFolder(parentPath, folderName)` (the bundle's IPC validator names the args **parentPath** then **folderName**), so the absolute path hit the UUID check and every call was rejected. COMPAT drift from 1.6259.1 → 1.14271.0; independent of the node-host fix (#142/#143).

## Fix

Rewrite the stub to the new contract:

- `create <parentPath>/<folderName>` on disk, **dedup on collision** (`name (1)`, `name (2)`, …) to match the native impl, return the created path.
- **Security:** the native macOS impl relies on the OS sandbox; Linux has none, so confine creation to the allowed roots (homedir, no `/tmp`) via the existing `requireAllowedPath` (same policy as `addFolderToSpace`), validating the **normalized** target so a `parentPath` with `..` segments can't escape.
- Removes the now-unused `isValidUUID`/`UUID_RE`.

## Verification

`node --check` passes. Exercised against the live store:

| case | result |
|------|--------|
| create folder | ✅ created, correct path |
| same name ×2 | ✅ `name (1)`, `name (2)` |
| parent `/tmp/...` | ✅ rejected |
| parent `/etc` (outside home) | ✅ rejected |
| name `../escape` | ✅ rejected |
| parent with `../../..` | ✅ rejected |
| empty name | ✅ rejected |
| null-byte name | ✅ rejected |
| relative parent | ✅ rejected |

### DIFF
--- a/stubs/cowork/spaces_store.js
+++ b/stubs/cowork/spaces_store.js
@@ -68,11 +68,6 @@ function createSpacesStore(options) {
     }
   }
 
-  const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
-  function isValidUUID(s) {
-    return typeof s === 'string' && UUID_RE.test(s);
-  }
-
   // Sanitize a value from the renderer to a plain string (max 10KB).
   // Rejects objects, arrays, and excessively large strings.
   function sanitizeString(v, maxLen) {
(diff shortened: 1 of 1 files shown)


johnzfitch:
Thanks @shawnyeager — your diagnosis (the asar 1.14271.0 `(parentPath, folderName)` contract change) and fix are spot on, and the verification table was great.

I've landed it in #146, which builds directly on your contract fix and adds:
- a Linux symlink-escape hardening — beyond the lexical `requireAllowedPath` check, it resolves the **nearest existing ancestor** through `realpath` and re-checks it, mirroring `addFolderToSpace`'s symlink defense (a lexical-only check can be bypassed by a symlinked ancestor like `~/Projects -> /etc`). The nearest-existing-ancestor resolution keeps your not-yet-created-parent behavior working.
- the first regression tests for `spaces_store` (contract, dedup, and the rejection/escape cases from your table, including a mutation-verified symlink test).

You're credited as co-author on the commit. Closing this in favor of #146 — much appreciated. 🙏

---

PR #152, opened 2026-07-08
fix: support Claude Desktop asar 1.19367.0 (split entry + code moved to chunk)

shawnyeager:
## Problem

Claude Desktop **1.19367.0** restructured its Vite main bundle in two ways that broke the Linux port across the board.

### 1. Split entry point → launch crash
The main process is now `index.pre.js` (stashes the `@sentry/electron/main` namespace on `globalThis.__sentryElectronMain`, then `require()`s `index.js`) + `index.js`. A sentry shim guard throws if `index.pre.js` didn't run first:

```
Error: sentryMainShim: globalThis.__sentryElectronMain is unset — index.pre.js
must stash its @sentry/electron/main namespace before index.js loads.
```

`frame-fix-entry.js` `require()`d `index.js` directly, skipping the stash → main process crashes on launch.

### 2. Main code moved into a chunk → every patch silently no-op'd
`index.js` is now a ~800-byte shim that `require()`s the real main code from `index.chunk-<hash>.js`. **Both** patch mechanisms targeted `.vite/build/index.js`, so both stopped applying:

- **`launch.sh`** (runtime seds): titlebar, file:// origin, `--effort xhigh→max`, macOS Handoff, `resourcesPath`, and the MCP node-host path. The last one surfaced as:
  ```
  MCP Filesystem: Node host not found at:
  …/.config/Claude/cowork-resources/app.asar/.vite/build/mcp-runtime/nodeHost.js
  ```
- **`enable-cowork.py`** (install-time, the actual Cowork *enablers*): the platform gate → `{status:"supported"}`, 711 IPC origin-validation guards (file:// exemption), `getHostPlatform()`, and return-style platform gates. Result: **zero** cowork markers in the packed app.

Four patterns additionally hardcoded minifier-assigned identifiers (`e`, `Ee`, `O`, `cA`, arg `i`, validators like `$m`) that rotate every build.

## Fix

- **`stubs/frame-fix/frame-fix-entry.js`** — load `index.pre.js` when present (it chains to `index.js`), fall back to `index.js` for older single-entry builds.
- **`launch.sh`** — discover the real code file(s) (`index.js` + the chunk it `require()`s) and apply every grep-guarded patch across them via a `patch_index` helper; generalize the four identifier-hardcoded patterns to `[A-Za-z0-9_$]+`.
- **`install.sh`** — `apply_patches` runs `enable-cowork.py` against `index.js` plus every chunk it `require()`s (same discovery as launch.sh).
- **`enable-cowork.py`** — generalize the IPC origin-guard regex: the sender-arg rotated `i`→`n` and some validators contain `$`, so match both with `[\w$]+` and reuse the captured arg in the exemption. Matches all 711 sites on 1.19367.0.

## Verification (end-to-end on asar 1.19367.0)

- App launches, main window opens, no sentry crash.
- MCP filesystem node-host resolves — error gone.
- All eight `launch.sh` patch passes report **applied** (previously zero reached the relocated code).
- All four `enable-cowork.py` patches apply to the chunk; patched chunk passes `node --check`; after repack the cowork markers are embedded in the running app.
- No `did not pass origin validation` throws, no `process is not defined`.

### DIFF
--- a/enable-cowork.py
+++ b/enable-cowork.py
@@ -146,8 +146,12 @@ def patch_host_platform(filepath):
 # add a file:// origin exemption at each call site: the original FUNC(i) still
 # validates non-file:// origins (e.g. would reject http://evil.com), so the
 # defense-in-depth layer is preserved for everything except our local renderer.
+# The validator name and the sender-arg name are both minified and rotate per
+# build (e.g. the arg was `i` in older builds and is `n` in 1.19367.0; validator
+# names like `$m` even contain `$`), so match both with [\w$]+ and reuse the
+# captured arg in the exemption rather than hardcoding it.
 IPC_ORIGIN_GUARD_RE = re.compile(
-    r'if\(!(\w+)\(i\)\)(throw new Error\(`[^`]*did not pass origin validation`\))'
+    r'if\(!([\w$]+)\(([\w$]+)\)\)(throw new Error\(`[^`]*did not pass origin validation`\))'
 )
 IPC_PATCH_MARKER = '/*cowork-ipc-patched*/'
 
@@ -162,10 +166,11 @@ def patch_ipc_origin_guards(filepath):
         print(f"  IPC origin guards: already patched")
         return True
 
-    # Replace if(!FUNC(i)) with if(!FUNC(i)&&!(i.senderFrame&&i.senderFrame.url&&i.senderFrame.url.startsWith("file://")))
+    # Replace if(!FUNC(ARG)) with
+    #   if(!FUNC(ARG)&&!(ARG.senderFrame&&ARG.senderFrame.url&&ARG.senderFrame.url.startsWith("file://")))
     # This lets file:// through while keeping the validator for everything else.
     new_content, count = IPC_ORIGIN_GUARD_RE.subn(
-        r'if(!\1(i)&&!(i.senderFrame&&i.senderFrame.url&&i.senderFrame.url.startsWith("file://")))\2',
+        r'if(!\1(\2)&&!(\2.senderFrame&&\2.senderFrame.url&&\2.senderFrame.url.startsWith("file://")))\3',
         content
     )
     if count == 0:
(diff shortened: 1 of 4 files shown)


vsem-azamat:
Verified this PR's split-entry/chunk changes locally against Claude Desktop `1.20186.0` on Arch, but there is an additional AUR-specific gap: the current AUR `PKGBUILD` invokes `enable-cowork.py` directly on `index.js` and launches the packed ASAR directly, so it bypasses both the `install.sh` chunk discovery and the `launch.sh` `package.json.main` rewrite in this PR.

The AUR recipe therefore needs two matching changes after this PR lands:

1. Patch `index.js` plus its referenced `index.chunk-*.js` targets, requiring at least one successful platform-gate match.
2. Rewrite the packed `package.json.main` to `frame-fix-entry.js`; otherwise the copied frame wrapper is never loaded and `1.20186.0` crashes on `systemPreferences.setUserDefault()`.

Full reproduction, proposed `PKGBUILD` change, and verification evidence are in #156. With both AUR changes applied locally: clean package build succeeds, 724 IPC guards patch, packaged markers are present, `package.json.main` is correct, 520/520 upstream tests pass, and runtime reaches `[cowork-startup] all rails engaged` with Cowork reported supported.

johnzfitch:
Thanks for this thorough write-up and fix, @shawnyeager — the split-entry + "code moved into a chunk" analysis was exactly right and drove a lot of the follow-up work.

The core has since landed in `master`:
- `index.pre.js` entry fix via #157 (commit `4636768`);
- chunk discovery + multi-target patching in `launch.sh`/`install.sh` via #158 (commits `898ac00`, `91b1a10`), with `enable-cowork.py`'s IPC origin-guard regex generalized to `[\w$]+` just as you had it here;
- the AUR `PKGBUILD` gap that @vsem-azamat surfaced in #156 was closed out separately in #159.

Since everything in this PR is now in `master`, I'm closing it as superseded — with thanks for the detailed diagnosis that got us there. 🙏

---

PR #155, opened 2026-07-09
fix(frame-fix): load index.pre.js so the sentry stash runs

### LINKED ISSUE #154: Asar 1.19367.0 Fix (for boot)
### Area

Launch (app won't start)

### What happened?

On asar 1.19367.0, the app crashes on launch with an uncaught exception in the main process:

A JavaScript error occurred in the main process
Uncaught Exception:
Error: sentryMainShim: globalThis.__sentryElectronMain is unset — index.pre.js must stash its @sentry/electron/main namespace before index.js loads. Check the entry detection in vite.main.config.mts and the stash in src/main/index.pre.ts.
    at tur (.../app.asar/.vite/build/index.chunk-<hash>.js:1:5526)
    ...
    at Module.require (.../app.asar/frame-fix-wrapper.js:1320:34)

Expected: app launches to the main window.
Actual: main process throws during load; no window.

Root cause. This newer build splits the Electron main entry into two files:

- .vite/build/index.pre.js — the real entry. It stashes the @sentry/electron/main namespace onto globalThis.__sentryElectronMain, then require("./index.js") itself.
- .vite/build/index.js — now contains a "sentryMainShim" that throws immediately if __sentryElectronMain is unset.

frame-fix-entry.js requires ./.vite/build/index.js directly, so index.pre.js never runs, the stash never happens, and the shim throws. (Older asars had no index.pre.js, so requiring index.js directly was fine — that's why this only shows up on newer builds.)

Proposed fix. Point the entry at index.pre.js instead. It stashes Sentry and then loads the same index.js (which still carries the yukonSilver/cowork patches), so nothing is lost:

  // stubs/frame-fix/frame-fix-entry.js
  require('./frame-fix-wrapper.js');
- require('./.vite/build/index.js');
+ // index.pre.js is the real entry: it stashes globalThis.__sentryElectronMain
+ // then require()s ./index.js itself (patches still apply). Loading index.js
+ // directly skips the stash and trips the sentryMainShim on newer asars.
+ require('./.vite/build/index.pre.js');

Confirmed working end-to-end after this change (main window loads, session spawn works) on asar 1.19367.0. Might be worth a fallback in enable-cowork.py/launch.sh: prefer index.pre.js when it exists, else index.js, to stay compatible with both old and new builds.

Secondary (UX) note. Because the crash is an uncaught main-process exception, Electron leaves the process alive showing only the error dialog. That stale process keeps the single-instance lock, so subsequent launches silently [SingleInstance] Another instance is running, quitting with no window and no obvious error. Recovery today is manual:

pkill -f 'app\.asar'
rm -f ~/.config/Claude/Singleton{Lock,Socket,Cookie}

Might be worth having the launcher detect/clear a stale lock, or having the wrapper app.exit(1) on fatal load errors so it doesn't linger holding the lock.

### Steps to reproduce

1. Install/update to Claude Desktop asar 1.19367.0 (claude-desktop --update, accept the newer-than-tested prompt).
2. Run ./launch.sh (or claude-desktop).
3. App crashes on launch; error dialog shows the sentryMainShim exception. No window.
4. Relaunching does nothing (stale instance holds the single-instance lock).

### Relevant log output

```shell
# ~/.local/state/claude-cowork/logs/startup.log
[Frame Fix] Wrapper v2.5 loaded
[Frame Fix] Intercepting electron module
[Frame Fix] BrowserWindow patched for native Linux decorations
...
App threw an error during load
Error: sentryMainShim: globalThis.__sentryElectronMain is unset — index.pre.js must stash its @sentry/electron/main namespace before index.js loads. ...
    at Module.require (.../app.asar/frame-fix-wrapper.js:1320:34)
A JavaScript error occurred in the main process

# on a second launch attempt:
[SingleInstance] Another instance is running, quitting
```

Arch / CachyOS — KDE Plasma (KWin), Wayland

057a2eb

1.19367.0 (last tested per COMPAT.md: 1.6259.1)

### Checklist

- [x] I checked the log file and redacted any tokens before pasting
- [x] I searched existing issues and this isn't a duplicate

Crono141:
## Problem

Fixes #154. On newer Claude Desktop asars (observed on **1.19367.0**) the app crashes on launch with:

```
Error: sentryMainShim: globalThis.__sentryElectronMain is unset — index.pre.js must stash its @sentry/electron/main namespace before index.js loads.
```

Because it's an uncaught main-process exception, Electron leaves the process alive showing only the error dialog, which then holds the single-instance lock and makes subsequent launches silently `[SingleInstance] Another instance is running, quitting` with no window.

## Root cause

These builds split the Electron main entry into two files:

- `.vite/build/index.pre.js` — the **real** entry. It stashes the `@sentry/electron/main` namespace onto `globalThis.__sentryElectronMain`, then `require("./index.js")` itself.
- `.vite/build/index.js` — carries the yukonSilver/cowork patches, but now begins with a `sentryMainShim` that **throws** if `__sentryElectronMain` is unset.

`frame-fix-entry.js` required `./.vite/build/index.js` directly, so `index.pre.js` never ran, the stash never happened, and the shim threw. Older asars (≤ the 1.6259.1 baseline) had no `index.pre.js`, which is why requiring `index.js` directly worked before.

## Fix

Load `index.pre.js` when it exists (it stashes sentry and then pulls in the still-patched `index.js`), and fall back to `index.js` on older asars that don't have it. Runtime `fs.existsSync` detection keeps **both** layouts working — no version assumption baked in.

## Testing

- asar **1.19367.0**, Electron 43.1.0, Arch/CachyOS, KDE Plasma (KWin) Wayland: app now launches to the main window, Cowork session spawn works, and `startup.log` shows **zero** `sentryMainShim` errors.
- `node --check stubs/frame-fix/frame-fix-entry.js` passes.
- Verified the detection selects `index.pre.js` on this install and would select `index.js` when absent.

## Out of scope

The secondary UX issue noted in #154 — Electron lingering after a fatal load error and holding the single-instance lock — is left for a separate change. This PR removes the crash that triggers it.

### DIFF
--- a/stubs/frame-fix/frame-fix-entry.js
+++ b/stubs/frame-fix/frame-fix-entry.js
@@ -1,4 +1,28 @@
 // Load frame fix first
 require('./frame-fix-wrapper.js');
-// Then load patched main (index.js has yukonSilver patches)
-require('./.vite/build/index.js');
+
+// Then load the real main entry.
+//
+// Newer asars (observed on 1.19367.0) split the Electron main entry in two:
+//   .vite/build/index.pre.js  — the true entry. It stashes the
+//     @sentry/electron/main namespace onto globalThis.__sentryElectronMain
+//     and then require()s ./index.js itself.
+//   .vite/build/index.js      — carries the yukonSilver/cowork patches, but
+//     now begins with a "sentryMainShim" that THROWS if __sentryElectronMain
+//     is unset.
+// Loading index.js directly skips the stash and crashes on startup with
+// "sentryMainShim: globalThis.__sentryElectronMain is unset". Loading
+// index.pre.js runs the stash and then pulls in the (patched) index.js, so
+// the patches still apply.
+//
+// Older asars (<= the 1.6259.1 baseline) have no index.pre.js, so fall back
+// to index.js there. Detect at runtime rather than assuming either layout.
+const fs = require('node:fs');
+const path = require('node:path');
+
+const preEntry = path.join(__dirname, '.vite', 'build', 'index.pre.js');
+if (fs.existsSync(preEntry)) {
+  require('./.vite/build/index.pre.js');
+} else {
+  require('./.vite/build/index.js');
+}


johnzfitch:
Thanks for this, @Crono141 — your diagnosis of the split `index.pre.js` entry and the sentry-shim stash was spot-on, and it directly drove the fix for #154.

The same change has since landed in `master` via #157 (commit `4636768`, "fix(frame-fix): load index.pre.js so the sentry stash runs (#154)"), using the same runtime `fs.existsSync(index.pre.js)` detection you proposed here so both asar layouts keep working. Since it's already in `master`, I'm closing this as superseded — but credit for identifying and fixing the crash is yours. 🙏

---

Crono141:
I cannot take credit for Claude"s work. I merely directed it to fix what I could not.

PR #160, opened 2026-07-21
fix(cowork): CoworkSpaces Linux stub fixes — live-refresh + asar 1.22209.x file-op contract

davide-perego-runtime:
Four related fixes to the Linux CoworkSpaces / native stubs. The first is version-independent; the rest are compat fixes for the split-entry asar **1.22209.x**, where the CoworkSpaces IPC contract and native file API changed.

## 1. Live-refresh: emit `onSpaceEvent`
On Linux, Spaces did not refresh live — create/archive/delete/add-folder persisted to `spaces.json` but the renderer was never notified, so changes only appeared after an app restart. macOS pushes these events from its native module; on Linux `onSpaceEvent` was a no-op.

- `spaces_store.js`: injected `emit`, called after every successful mutation with `{ type: 'created'|'updated'|'deleted', space }`.
- `frame-fix-wrapper.js`: broadcasts via `webContents.send` to the `CoworkSpaces … onSpaceEvent` eIPC channel; the channel UUID is **captured dynamically** at handler registration (self-heals across asar builds — verified live when the UUID changed to `960045f6…` on 1.22209.x), seeded with the tested UUID for the cold path.
- `ipc_overrides.js`: fallback store reuses the transport via `global.__coworkEmitSpaceEvent`.

## 2. File-op arg contract `(path)` -> `(spaceId, path)`
asar 1.22209.x prepends a `spaceId`, moving the real path to argument position 1 for `listFolderContents` / `readFileContents` / `openFile`. The stubs read position 0 as the path, got the UUID, and folder browsing silently returned empty (`listFolderContents BLOCKED: <uuid>`).

- `spaces_store.js`: `pickPathArg(a,b)` prefers the trailing path, accepting **both** shapes (rollback-safe). Wiring in `frame-fix-wrapper.js` / `ipc_overrides.js` forwards both args. (User-confirmed: folder contents now display.)

## 3. Auto-memory dir browsing
The UI lists a space's internal auto-memory dir (`localAgentRoot/.../spaces/<id>/memory`) — not a registered project folder, so it was blocked. `resolveWithinSpaceMemory` allows that subtree for read only; the rest of `localAgentRoot` stays off-limits.

## 4. Native safe-fs containment (`@ant/claude-native.openRootDir`)
1.22209.x routes artifact/upload/attachment/transcript writes through a native safe-fs API (`openRootDir` + `mkdir/rename/unlink/openBeneath`, openat2 RESOLVE_BENEATH on macOS). Linux had no binding → `openRootDir is not a function` → recurring `UnsafeRootError`, breaking those writes (app didn't crash).

- `stubs/@ant/claude-native/safe_fs.js`: Linux implementation. Containment approximated with realpath checks (lexical + nearest-existing-ancestor), **fail-closed** on separator/`..`/NUL/symlink-escape. Byte I/O delegated to Node `fs.promises` FileHandle (`read(buf,off,len,pos)`/`write`/`stat`/`close` match the caller).
- `launch.sh` / `install.sh`: sync sibling helper modules (`safe_fs.js`), not just `index.js`, so the now-multi-file native stub deploys and repacks.

## Testing
`spaces_store.test.cjs` (16) + `native_safe_fs.test.cjs` (3): event emission, both file-op contracts, scoped auto-memory access, safe-fs round-trip + fail-closed containment (separators, `..`, symlink escape). Full suite green apart from one pre-existing environmental failure (`getAllSpaces returns empty array`, unrelated).


johnzfitch:
Reviewed — this is good work, and all four commits are carried into #162 with your authorship intact. The contract archaeology on 1.22209.x is the useful part: accepting both `(path)` and `(spaceId, path)` so a rollback still works, and capturing the eIPC UUID dynamically instead of pinning it, are both the right calls.

One real finding in `safe_fs.js`, fixed on top rather than sent back to you.

**`openBeneath` could be escaped via a dangling symlink.** `resolveBeneath` proves containment by realpathing the nearest *existing* ancestor, and `existsSync()` returns false for a broken link — so the walk steps straight past the link to the legitimate parent, the check passes, and `open()` with a creating flag then follows the link and writes at its target outside the root. Reproduced against your branch:

```js
fs.symlinkSync(path.join(outside, 'pwned'), path.join(root, 'dangle'));
const fh = await safeFs.openBeneath(h, ['dangle'], 'w');
await fh.write(Buffer.from('ESCAPED'), 0, 7, 0);
// -> outside/pwned now exists, containing "ESCAPED"
```

Your symlink test passes only because it links to a directory that *exists*, which `existsSync()` does see. The dangling case is the gap — and it's the reachable one, since the roots here hold agent-written artifacts and uploads.

The fix opens the final component with `O_NOFOLLOW` so the kernel refuses to follow it. On `ELOOP` the link is resolved and re-checked against the root, which preserves the `RESOLVE_BENEATH` behavior you were approximating (links that stay beneath the root keep working); a dangling link can't be resolved and is denied. That needed string open flags mapped to numeric bits to OR the flag in, so unrecognized flag strings are now rejected rather than silently opened without it.

Two smaller notes:

- The live-event broadcast went to every `webContents`. The payload carries space names and registered folder paths, so it's now scoped to `window`/`browserView` — `<webview>` hosts embedded non-app content and shouldn't see it.
- `resolveWithinSpaceMemory` I looked at closely, since it does widen the read surface, and kept as-is. It's read-only, scoped to the `memory` subtree with the rest of `localAgentRoot` still blocked, `openFile` still requires a registered folder, and the realpath-before-prefix-check means a symlink inside the memory dir can't redirect out. Sound.

Everything else went in unchanged. `COMPAT.md` now records 1.22209.x as `[PARTIAL]`, crediting your live report — `LAST_TESTED_ASAR_VERSION` stays put until someone exercises it end-to-end on a GUI session.

Since the fix touches your new file, #162 supersedes this PR; closing it there rather than asking you for another round. Thanks for the thorough test coverage — it made the gap easy to isolate.

---

Issue #161, opened 2026-07-24
[BUG] "Couldn't link this session to a computer" — DeviceRegistry and BuddyBleTransport stubs not implemented


rgaspary:
### Area

Cowork session (spawn / spawn fails)

### What happened?

Every Cowork task session fails immediately with:

"Couldn't link this session to a computer, so attached folders can't be used. You can restart this task on your computer instead."

### Root cause

The startup log shows two unhandled IPC channels firing on every session attempt:

```bash
Error occurred in handler for '$eipc_message$_..._$_claude.web_$_DeviceRegistry_$_signCreateSessionBind':
Error: DeviceRegistry: device not registered (no row-PK for this account)

Error: No handler registered for '$eipc_message$_..._$_claude.buddy_$_BuddyBleTransport_$_reportState'
```
Neither `DeviceRegistry_$_signCreateSessionBind` nor `BuddyBleTransport_$_reportState` have stub handlers — they are absent from `ipc_overrides.js` and everywhere else in the cowork stubs. The `DeviceRegistry` failure is the direct cause of the session link error; `BuddyBleTransport` appears to be part of the same session startup handshake.

### What would fix it

Stub handlers for at minimum these IPC channels in `ipc_overrides.js`:

`DeviceRegistry_$_signCreateSessionBind` — needs to either implement real device registration with Anthropic's API or return a fake registration that satisfies the session bind check
`BuddyBleTransport_$_reportState` — likely needs a no-op stub to prevent the unhandled error from blocking session startup

### Steps to reproduce

1. Launch claude-desktop
2. Navigate to the Cowork tab
3. Start any task with an attached folder
4. Observe the "Couldn't link this session to a computer" banner immediately

### Relevant log output

```shell
### Relevant log output


[CCD] Initialized with version 2.1.217
Error occurred in handler for '$eipc_message$_..._$_claude.web_$_DeviceRegistry_$_signCreateSessionBind': Error: DeviceRegistry: device not registered (no row-PK for this account)
Error: No handler registered for '$eipc_message$_..._$_claude.buddy_$_BuddyBleTransport_$_reportState'
Uncaught (in promise) Error: Error invoking remote method '$eipc_message$_..._$_claude.buddy_$_BuddyBleTransport_$_reportState': Error: No handler registered


### Doctor output


19 passed  2 warnings  0 failed
WARN: Asar version: 1.24012.1 (newer than last tested 1.6259.1)
```

### Distro and desktop environment

[...]

Issue #165, opened 2026-08-01
No download URL for 1.19367.0, although it's the latest marked OK in COMPAT.md


elvanor:
### Area

Install / setup

### What happened?

A download URL should be provided in COMPAT.md, as it's not easy to find it.
The current latest is at https://downloads.claude.ai/releases/darwin/universal/1.19367.0/Claude-1a5be1fbf83d1832486e03a667557c18f0a0ec7a.dmg

I tried to update today but this was again an obstacle to overcome.

### Steps to reproduce

./install.sh
Read COMPAT.md

### Relevant log output

```shell

```

### Distro and desktop environment

Gentoo Linux

### claude-cowork-linux version / commit

0152733

### Checklist

- [ ] I ran `./install.sh --doctor` and included its output if relevant
- [ ] I checked the log file and redacted any tokens before pasting
- [ ] I searched existing issues and this isn't a duplicate

PR #167, opened 2026-08-07
fix: restore Cowork patching on asar 1.26832.0 (+ 3 version-independent bugs)

### LINKED ISSUE #166: asar 1.26832.0: every patch silently no-ops (platform gate not found)
On asar **1.26832.0** (current CDN latest), `install.sh` ends with:

```
[WARN] Cowork patch matched no target: the platform gate was not found in index.js or any index.chunk-*.js.
```

and **all 8 `launch.sh` patches no-op silently** — the launcher logs no `Patching ...` lines at all. `makepkg` fails the same way in `build()`. Cowork is completely unavailable.

Three independent causes, each of which alone breaks discovery or matching:

**1. A second chunk series.** Split-entry builds now emit `index2.chunk-<hash>.js` alongside `index.chunk-<hash>.js`, and the Cowork platform gate moved into the index2 series (`ke()` in `index2.chunk-CQIegP9t.js` here). Discovery matches `index\.chunk-` only, so the gate-bearing chunk is never passed to the patcher — 86 of 131 chunk refs seen.

**2. The minifier switched to backtick template literals.** `"darwin"` is now `` `darwin` ``, `"hidden"` is `` `hidden` ``, and so on, so every pattern anchored on a double-quoted literal misses. The gate is:

```js
function ke(){let t=process.platform;if(t!==`darwin`&&t!==`win32`)return{status:`unsupported`,...
```

which also uses `let` (patterns assume `const`). Related in the same build: a bare `throw Error(` where the IPC/host-platform patterns require `new`, and minified values that gained dots and negations (`r.default`, `!1`) that `[A-Za-z0-9_$]+` won't match.

**3. Chunks require() each other transitively.** Following only the shim's requires reaches 131 of 333 chunks. The `--effort` builder lives in `index2.chunk-Cqfh0Vpp.js`, which `index.js` never names, so the `xhigh -> max` remap silently does nothing even once (1) and (2) are fixed.

While fixing the above I hit two bugs that are **not** version-specific:

**4. `enable-cowork.py` gates the auxiliary passes on the platform gate.** `__main__` only runs `patch_host_platform` / `patch_ipc_origin_guards` / `patch_platform_return_gates` if `patch_file()` succeeded *on that same file*. That held when everything lived in one `index.js`, but on any split bundle the gate is in one chunk while the IPC origin guards are spread across ~10 — so those passes are skipped in every chunk but one. Here it meant 10 files that should have been patched got nothing.

**5. `--doctor` false negative.** It greps only `index.js` for `cowork-patched`, but on a split build the marker lives in a chunk, so a correctly patched install reports `Cowork patch: not applied (run install.sh to apply)`.

And one unrelated bug found on the way:

**6. The `claude://` fast path is dead code.** `install_stubs()` copies `protocol-forwarder.js` to `linux-app-extracted/`, but the generated launcher looks for it at `$COWORK_DIR/protocol-forwarder.js`. It never resolves, so every OAuth callback falls through to a full `launch.sh` run — re-patch, repack app.asar, second Electron. Mid-login the app visibly closes and reopens.

**Environment:** Manjaro (Arch), Cinnamon/X11, Electron 43.1.1, Node 26, asar 1.26832.0.

PR follows. With all six fixed: app launches, all 8 launch.sh patches apply, `[Cowork] Linux support enabled`, 21 CoworkSpaces handlers registered, all four `cowork-startup` rails ok, `--doctor` 20/1/0, sign-in works and the Cowork surface loads. `tests/test-cowork-patch.sh` 33/33.

I did not re-test a full Cowork session end-to-end, so I can't say whether #161's device-binding failure still applies on this build — I saw no banner, but I only verified the surface loads.

chefboyrdave21:
Fixes #166.

On asar 1.26832.0 every patch silently no-ops and Cowork is unavailable. Three regressions cause it; while fixing them I hit three more bugs that are not version-specific. Each pattern change is a **superset** of the old one, so older builds and the existing fixtures keep matching.

### The 1.26832.0 breakage

| # | Cause | Fix |
|---|---|---|
| 1 | Gate moved into a new `index2.chunk-*.js` series; discovery matched `index\.chunk-` only | match `index[0-9]*\.chunk-` |
| 2 | Minifier switched to backtick literals (`` `darwin` ``), plus `let` vs `const`, bare `throw Error(`, and dotted/negated values (`r.default`, `!1`) | quote-agnostic patterns, widened char classes |
| 3 | Chunks require() each other transitively; the shim names 131 of 333 | discover by globbing the build dir |

The gate on this build:

```js
function ke(){let t=process.platform;if(t!==`darwin`&&t!==`win32`)return{status:`unsupported`,...
```

(3) is what makes the `--effort xhigh -> max` remap work: it lives in `index2.chunk-Cqfh0Vpp.js`, which `index.js` never references.

### Bugs that aren't version-specific

- **`enable-cowork.py` gated the auxiliary passes on the platform gate.** They only ran if the gate was found *in the same file*, so on any split bundle the IPC origin-guard / host-platform / return-gate passes were skipped in every chunk but one. Here that was **10 files** that should have been patched and weren't. They're each marker-guarded already, so running them unconditionally is safe; the exit code still reports the gate alone, which is what `install.sh` keys off.
- **`--doctor` false negative** — grepped only `index.js` for the marker, so a correctly patched split build reported "not applied" and sent users into a pointless reinstall.
- **The `claude://` fast path was dead code** — `install_stubs()` writes `protocol-forwarder.js` into `linux-app-extracted/`, the launcher looked only at the install-dir root. Every OAuth callback fell through to a full `launch.sh` run (re-patch + repack + second Electron), so the app visibly closes and reopens mid-login.

Also included: `install.sh` honours `CLAUDE_REPO_URL` / `CLAUDE_REPO_REF` / `CLAUDE_INSTALL_DIR`. Without it there's no way to install a fork end-to-end, which is what this work needed; defaults are unchanged.

### Verification

Manjaro (Arch), Cinnamon/X11, Electron 43.1.1, Node 26, asar 1.26832.0:

- All 8 `launch.sh` patches apply (previously zero)
- `[Cowork] Linux support enabled - VM will be emulated`; 21 CoworkSpaces + 194 fallback handlers registered
- All four `cowork-startup` rails `ok:true`
- `install.sh --doctor`: 20 passed, 1 warning, 0 failed
- Sign-in completes and the Cowork surface loads
- `tests/test-cowork-patch.sh`: **33/33**

The one test change is `tests/test-cowork-patch.sh`: the PKGBUILD discovery guard asserted the literal regex string rather than the behaviour, so it now accepts either mechanism.

Two things worth flagging:

- `tests/test-install-paths.sh` reports 4 failures on this branch — but **byte-identical failures on unmodified `master`**, so they're pre-existing here, not introduced. Two of them (`makepkg failed`, `Docker install: PATCH_MISSING`) are *this very bug*: both stages clone `master` for their scripts, so they can't pass from a fork. I verified the PKGBUILD path separately by pointing `source=` at the fork branch and running `makepkg`.
- COMPAT.md records 1.26832.0 as `[PARTIAL]`, not `[OK]`: I verified launch, patching and the Cowork surface, but did not drive a session end-to-end, so I can't speak to whether #161's device-binding failure still applies here. I saw no banner.

Written with an AI coding agent (per CONTRIBUTING). Every change was reviewed and verified against the live 1.26832.0 bundle rather than assumed — including one approach I had to throw away after proving it didn't work. No agent instruction files are added.


chefboyrdave21:
Follow-up on the PKGBUILD verification mentioned above — now done end to end.

Pointed `source=` at this fork's branch and ran `makepkg -sf`:

```
==> Finished making: claude-cowork-linux 1.26832.0-1
```

Exit 0, `claude-cowork-linux-1.26832.0-1-x86_64.pkg.tar.xz` (102 MB). Confirmed the gate is actually patched in the packaged asar rather than just assuming the build's exit code meant success:

```
$ asar extract-file usr/lib/claude-cowork/app.asar .vite/build/index2.chunk-CQIegP9t.js
$ grep -o 'function ke(){return{status:"supported"}}/\*cowork-patched\*/' ...
function ke(){return{status:"supported"}}/*cowork-patched*/
```

Note the pkgver comes out as `1.26832.0-1`, versus `1.1.4010-10` currently on AUR — so this also unblocks the AUR package, which still fails in `build()` on any current CDN build for the reason in #166.

For anyone hitting this before the AUR package is refreshed: `makepkg` against master will keep failing, because the PKGBUILD clones `master` for its scripts. Building from a branch with these fixes is the workaround.

chefboyrdave21:
Updated COMPAT.md for 1.26832.0 from `[PARTIAL]` to `[OK]`.

The PARTIAL note called out one gap — nothing exercised past sign-in. That's now closed: sign-in through the `claude://` callback works, and live Cowork sessions are working in day-to-day use on the account that hit the original breakage. The build has also now been produced both ways, via `install.sh` and via `makepkg` from the PKGBUILD.

Two notes I've folded into the table entry rather than overstate:

- **#161 is left explicitly unconfirmed, not fixed.** The device-binding failure did not appear here, but nobody deliberately drove that path, so it shouldn't be read as resolved on this build.
- **The UI moved upstream between 1.1.x and 1.26832.0**: Cowork is no longer a top-level tab, it's a `Chat`/`Cowork` mode toggle on the composer. Worth recording because it's a cheap signal that the platform gate is patched — stock upstream on Linux shows Chat only, so if the toggle is missing, `ke()` didn't get patched.

shawnyeager:
Applied the `launch.sh` + `enable-cowork.py` changes from this branch on Arch against the same 1.26832.0 bundle (same chunk hashes — gate in `index2.chunk-CQIegP9t.js`) and they work as described: gate patched, MCP node-host paths rewritten, Filesystem MCP starts and connects.

One site survives the widened MCP pattern, though. The `shellPathWorker.js` lookup in `index.chunk-CDtvTf2K.js` has no `isPackaged` guard at all:

```js
function v(){return o.default.join(process.resourcesPath,`app.asar`,`.vite`,`build`,`shell-path-worker`,`shellPathWorker.js`)}
```

Since the new pattern still requires `\.app\.isPackaged\?`, this join keeps resolving through the overridden `process.resourcesPath` (`~/.config/Claude/cowork-resources`), which has no `app.asar` — so login-shell PATH priming breaks on Linux. A catch-all pass placed after the guarded one covers it; the main-process chunks are CJS with `require` in scope, so:

```sh
patch_index "Patching unconditional resourcesPath+app.asar joins to use getAppPath()..." \
  'process\.resourcesPath,["`]app\.asar["`]' \
  's/process\.resourcesPath,["`]app\.asar["`]/require("electron").app.getAppPath()/g'
```

Running it after the guarded pass means it only touches sites that pass missed. Verified locally on top of this branch: the packed asar ends up with `nodeHost.js`/`directMcpHost.js` on `getAppPath()` via your pattern and `shellPathWorker.js` via the catch-all, no `resourcesPath,app.asar` join remains in any `index*.chunk-*.js`, and on relaunch the Filesystem MCP server logs "Server started and connected successfully" where it previously threw "Node host not found".

johnzfitch (commented):
Reviewed, and I reproduced the core claim rather than taking the report on faith. This is a good diagnosis — three independent causes is exactly the kind of thing that gets half-fixed, and splitting them out made it checkable.

## What I verified

I wrote a fixture shaped like 1.26832.0 — backtick literals, `let`, bare `throw Error(`, values that gained dots and negations (`r.default`, `!1`) — and ran both trees against it.

**master:** `enable-cowork.py` reports "Platform-gate function not found", and *zero* of the `patch_index` passes fire. Nothing logs, nothing changes. That matches #166 exactly, including why it's silent.

**This branch:** gate patched, host-platform throw rewritten, IPC origin guards patched, return gates neutralized, and all five `patch_index` passes fire:

```
function ke(){return{status:"supported"}}/*cowork-patched*/
```

`node --check` clean afterward. I also merged the branch onto current master and re-ran: `tests/test-cowork-patch.sh` 35/35, fixture still fully patched.

Two design calls I want to name because they're easy to get wrong and you got them right: keeping the exit code gate-only while running the auxiliary passes unconditionally preserves the contract `install.sh` keys off, and writing each pattern as a **superset** rather than swapping `"` for `` ` `` means the older fixtures keep matching. The `enable-cowork.py` `__main__` bug (4) is the best find in here — silently skipping ten files is the kind of thing that only shows up as a vague misbehaviour much later.

## Before this lands

**1. It needs a rebase — it doesn't currently merge.** The branch is based on `0152733`; master is now `7115f3e` (#168). Only `COMPAT.md` conflicts: #168 added a `1.24012.9` row to the same table this adds `1.26832.0` to. `launch.sh`, `PKGBUILD`, `enable-cowork.py`, `install.sh` and the test all auto-merge. I checked the one interaction I was worried about — #168's `launch.sh` change is comment-only (the disclaimer-wrap warning), so there's no pattern collision.

**2. Add a pinning row for 1.26832.0.** This bumps `LAST_TESTED_ASAR_VERSION` without recording where that archive came from. That's #165: the CDN URL embeds a per-release hash, so a version recorded as tested with no URL is a version nobody can obtain, and `install.sh`'s "install the tested version" path has nothing to print. #169 (just opened) fixes that for 1.19367.0 and adds `tests/test-compat-pins.sh`, which asserts the pin exists — so with both merged, the suite will fail until this row is added. You downloaded the archive, so this is a URL and one `sha256sum`.

**3. The quote-style flip has no test.** This is my main substantive concern. The regression that motivated the PR — every double-quote-anchored pattern missing — is invisible to the suite, because `write_patch_fixture` emits one double-quoted fixture and every assertion reads back double-quoted output. Someone could re-narrow `["`]` to `"` tomorrow and get 35/35 green. Given that the failure mode here is *silent* (patches no-op, install reports success), that gap is worth closing in the same PR that widens the patterns. I have the backtick fixture from my verification above and I'm happy to hand it over or land it as a follow-up — whichever you prefer, but I'd rather it not be left as "someone should".

**4. `discover_chunks()` in the test suite is now stale.** Section 3 defines it as "The exact discovery used by install.sh apply_patches and launch.sh" and still greps `index\.chunk-` out of the shim. After this PR both use `find … -name 'index*.chunk-*.js'`. The test still passes, but it's now asserting a mechanism the shipped code doesn't have — the same drift the PKGBUILD guard hit, which you fixed one section further down.

**5. @shawnyeager's `shellPathWorker.js` finding is real** — I confirmed it on the fixture. The unguarded join survives this branch's widened pattern, because the pattern still requires `\.app\.isPackaged\?`:

```js
o.default.join(process.resourcesPath,`app.asar`,`.vite`,`build`,`shell-path-worker`,`shellPathWorker.js`)
```

The catch-all-after-the-guarded-pass ordering in that comment is the right shape. Worth folding in here rather than trailing behind, since it's the same class of bug and the same file.

## Two notes, not blockers

**Blast radius of the unconditional passes.** Running them everywhere is correct, but note `patch_host_platform` is the one pass with no marker, and it `sub`s globally — it now rewrites `throw … Unsupported platform …` → `return"darwin-x64"` across every chunk instead of one. Right for `getHostPlatform()`; for any *other* function carrying that string it silently returns a platform name instead of throwing. Nothing in the current build appears to hit that, and I'm not asking you to narrow it. But a `node --check` sweep over patched chunks in `install.sh`/`launch.sh` would be a cheap guard now that we're rewriting ~300 files instead of 2, and it would have caught this class of thing at install time rather than at launch.

**`tests/test-install-paths.sh` — one correction.** You attributed `Docker install: PATCH_MISSING` to the stage cloning `master`, but that stage does `cp -r /mnt/repo` into the install dir, so it runs *your* scripts. The actual cause is line 409:

```sh
grep -q cowork-patched …/linux-app-extracted/.vite/build/index.js
```

which is the identical index.js-only marker grep this PR fixes in `--doctor` (bug 5). So that assertion will keep failing on any split build even after this merges. Same one-line fix as the doctor one; you may as well take it while you're in there.

`CLAUDE_REPO_URL` / `CLAUDE_REPO_REF` / `CLAUDE_INSTALL_DIR` are fine by me — I checked that `COWORK_DIR` in the generated launcher derives from `INSTALL_DIR`, so an override propagates rather than half-applying. Worth a line in CONTRIBUTING so it's a documented testing surface rather than folklore.

Thanks for doing this against the live bundle and for saying plainly which parts you didn't exercise — the note that #161 is unconfirmed rather than fixed on this build is exactly the right call, and I've kept that distinction.

---

chefboyrdave21:
Thanks for reproducing it on your own fixture rather than taking the report on faith — and @shawnyeager for the `shellPathWorker.js` find. All five blockers and both notes are addressed; branch force-pushed.

**First, your correction on `test-install-paths.sh` — you're right and I was wrong.** I claimed `Docker install: PATCH_MISSING` was the stage cloning `master`. It isn't; the stage `cp -r /mnt/repo` and runs our scripts. The actual cause is exactly what you said: line 409 is the same index.js-only marker grep this PR fixes in `--doctor`. Fixed there too, same one-liner. I should have read the stage instead of pattern-matching it to the `makepkg` failure next to it.

### The five

**1. Rebased** onto `7115f3e`. Only `COMPAT.md` conflicted, as you predicted; the `launch.sh` interaction you flagged was indeed comment-only. Merges clean now.

**2. Pinning row added.** SHA-256 is computed from the archive this branch was actually built and verified against, not copied out of `fetch-dmg.js`:

```
6b0f2c51c5e1c3f6db3885233ad96e48f92e61438b0bd9892c69f2ea11c54950
```

URL re-checked live: HTTP 200, `application/zip`, 350467183 bytes.

**3. Backtick fixture added** (new section 6) — I'd rather land it here than leave it as "someone should," so I wrote one rather than take yours; shout if you prefer yours. It covers backticks, `let`, bare `throw Error(`, and the dotted/negated values, across both `enable-cowork.py` and the `launch.sh` passes.

I checked it actually has teeth by reintroducing the regression — re-narrowing `["`]` back to `"` in `launch.sh` and `enable-cowork.py`:

```
49 passed, 6 failed
```

That surfaced a gap worth naming: with the narrowing in place, the **Python gate assertions still passed**, because the `ke()` fixture matches the exact `KNOWN_PATTERNS` entry and short-circuits `PLATFORM_GATE_RE` entirely. Since minified names rotate every build, the regex fallback is what actually carries the *next* one, so it needs its own coverage. Added a second fixture with a gate name in no known-pattern list (`zQ7x`) that forces the fallback path.

**4. `discover_chunks()` un-drifted.** Now takes the build dir and globs, matching the shipped code, with cases for the index2 series, a transitively-required chunk the shim never names, and non-chunk files in the same dir.

**5. `shellPathWorker.js` folded in**, with the catch-all-after-guarded ordering. Verified on the live 1.26832.0 tree: after the full pass, zero `process.resourcesPath,`app.asar`` joins remain in any `index*.chunk-*.js`, and the site reads

```js
o.default.join(require("electron").app.getAppPath(),`.vite`,`build`,`shell-path-worker`,`shellPathWorker.js`)
```

### Both notes

**`node --check` sweep added** before repack, over the discovered targets. Warns rather than aborts, so a pre-existing syntax error in a chunk we never touched can't block an otherwise-fine launch. Clean on the live tree. Agreed on the markerless global `patch_host_platform` — I left its scope alone.

**`CLAUDE_REPO_*` documented** in CONTRIBUTING, including the caveat that bit me: `PKGBUILD`'s `source=()` clones this repo by URL, so `makepkg` runs *master's* scripts even from a fork checkout. That, not the Docker stage, is why `makepkg failed` reproduces identically on both trees.

Suite: **58 passed, 0 failed.** Re-verified end to end on the live bundle after the rebase — all passes fire, `[Cowork] Linux support enabled`, all four rails ok, `node --check` clean, sign-in and Cowork surface working.

### One thing I noticed but did not touch

`PKGBUILD` runs 3 `patch_index` passes; `launch.sh` now runs 10. Since `/usr/bin/claude-cowork` execs electron directly on the packaged asar and never calls `launch.sh`, the packaged build appears to miss the resourcesPath, MCP node-host, `--effort`, Handoff and now the `shellPathWorker` passes — so an AUR user would still hit "Node host not found". That's pre-existing and orthogonal to #166, so I've deliberately left it out rather than grow this PR. Happy to open a separate issue, or fold it in here if you'd rather it travel with the catch-all.

johnzfitch:
Re-verified the updated branch (`8908fb0`) locally. All five blockers and both notes are addressed, and the two things I most wanted checked hold up.

**Suite on your branch: 58 passed, 0 failed.** Matches what you reported.

**The backtick fixture has teeth.** I reintroduced the regression independently — re-narrowed `["`]` to `"` across `launch.sh` and `enable-cowork.py` — and got **50 passed, 8 failed**. (You saw 49/6; the delta is just that my narrowing was cruder and caught the `mainView.js` guard too. Same conclusion: the fixture fails loudly when someone undoes the fix, which is exactly what it's for.)

The `zQ7x` fixture is the part I want to single out. You went looking for whether your own new test actually exercised the fallback, found it didn't — `ke()` short-circuits on `KNOWN_PATTERNS` — and added a second fixture for the path that carries the *next* build. That's a better catch than anything I raised, and it's the difference between a test that pins today's build and one that pins the mechanism.

**The #169 interaction is resolved.** I merged your branch against #169 and ran both suites:

```
tests/test-compat-pins.sh   11 passed, 0 failed
tests/test-cowork-patch.sh  58 passed, 0 failed
```

`LAST_TESTED_ASAR_VERSION=1.26832.0` now has its pinning row, so the assertion I warned about passes. One conflict on merge, in the pinning table — both sides append a row, resolution is to keep both. I'll handle that when landing; nothing for you to do.

Your `.zip` URL also exercises a branch of #169 I couldn't reach before (I only had a `.dmg` pinned), so the installer now emits a correct, complete instruction for your build:

```
curl -fL -o Claude-1.26832.0.zip "https://downloads.claude.ai/releases/darwin/universal/1.26832.0/Claude-056ee2be…zip"
echo "6b0f2c51…c54950  Claude-1.26832.0.zip" | sha256sum -c
```

Thanks also for the plain correction on `test-install-paths.sh` rather than quietly fixing it — and for fixing line 409 while you were there.

**On the `PKGBUILD` gap you noticed and left alone:** you called that right. `PKGBUILD` running 3 passes where `launch.sh` runs 10 is a real bug — AUR users would hit "Node host not found" — but it's pre-existing, orthogonal to #166, and this PR is already large. Please open it as a separate issue rather than folding it in; I'd rather land the #166 fix and track that on its own, and the packaged-vs-launcher divergence deserves its own discussion (the right fix is probably to stop maintaining two patch lists, not to copy seven more passes into the recipe).

No further changes requested from me.

---

chefboyrdave21:
Opened as #170, kept out of this PR as you asked.

Included the measurements rather than just the pass-count diff — the package I measured was built from a branch carrying **all** of this PR's `PKGBUILD` fixes, so it rules out "the recipe didn't get the fix" and isolates it to the recipe never having had those six passes:

```
packaged (AUR path)   4 unpatched process.resourcesPath,`app.asar` joins,  0 xhigh remap
launch.sh-patched     0                                                    1
```

I also agree with your read on the fix and said so there: copying six passes into the recipe just recreates the drift a build later. Sketched three shapes for a single shared list, with sourcing a shared `patch-index.sh` from both `launch.sh` and `build()` as the one I'd pick — it also collapses sections 4 and 5 of the test suite into one, which makes the divergence unrepresentable instead of merely tested for. Left the call to you rather than guessing at the refactor; happy to take it once you've picked.

Thanks for the thorough review on this one. The `zQ7x` gap only turned up because your point 3 made me check whether my own new test actually exercised the fallback, so that catch is downstream of yours.

johnzfitch:
Heads up: #169 just merged (`c6a33a3`), so GitHub now flags this as conflicting. That's my doing, not a regression on your side — both branches append a row to the pinning table, which is the conflict I said I'd handle at merge time.

**Please don't rebase for it.** The resolution is to keep both rows, and I already have it staged and tested locally: 11/11 on `test-compat-pins.sh`, 58/58 on `test-cowork-patch.sh`, with `LAST_TESTED_ASAR_VERSION=1.26832.0` resolving to your `.zip` URL and checksum. If you have maintainer edits enabled on the branch I'm happy to push the resolution to it; otherwise it gets resolved when this lands.

Nothing else changed for you — this is still verified and ready from my side.

---

johnzfitch:
Ran an adversarial security review over this before landing, staged as a merge onto master so the diff under review was the real pre-landing state. **No exploitable findings.** Posting the result because one part of it is worth you and @shawnyeager knowing: the `shellPathWorker` catch-all is a security fix, and neither of you framed it as one.

### The catch-all closes a plant-and-execute path

`stubs/frame-fix/frame-fix-wrapper.js:12` redefines `process.resourcesPath` to `$XDG_CONFIG_HOME/Claude/cowork-resources` — a **user-writable** directory. So on master today the unguarded join resolves to:

```
~/.config/Claude/cowork-resources/app.asar/.vite/build/shell-path-worker/shellPathWorker.js
```

Electron forks that as a worker, and nothing creates it. Anything able to write under `~/.config` can plant a file at that path and have it executed **in the Electron main process** on next launch. That matters more here than the usual same-uid shrug, because this project runs an agent that writes files as the user and has a whole admission layer (`exec_capability_registry.js`, `user_mcp_allowlist.js`) devoted to constraining what it may spawn — main-process code execution sits on the wrong side of that boundary.

After this PR it resolves through `app.getAppPath()`, inside the packed asar. So @shawnyeager's report was a security fix wearing a "PATH priming is broken" hat.

I checked whether the same shape lurks at the sites the catch-all *doesn't* reach. Of 12 `process.resourcesPath` uses in the bundled `index.js`, only 2 carry the `,"app.asar"` join — the other being the `isPackaged`-guarded `nodeHost.js`, already rewritten by the preceding pass. The rest load data, not code. The one exception, `Helpers/disclaimer`, **is** executed, but the wrapper already guards it: it only overrides `resourcesPath` when no root-owned (`uid === 0`) system disclaimer exists. That case was thought about; `shellPathWorker` was the one that wasn't.

### The other three areas, all cleared

- **`CLAUDE_REPO_URL` / `_REF` / `_INSTALL_DIR`** — read in exactly two places (`install.sh`, CONTRIBUTING). Nothing sets them from a config file, desktop entry, systemd unit, `.env`, `source`, or `export $(…)`, and `main()` refuses to run as root, so they never reach a privileged operation. `git clone` with a hostile URL and `CLAUDE_INSTALL_DIR` containing `$(…)` are both real primitives, but both require control of the invoking user's own environment.
- **Widened `file://` IPC exemption** — not a new exemption, a restored one. On single-file bundles all 341 origin guards in the repo's bundled `index.js` were already patched; the `patch_file` gate only ever *narrowed* this on split builds. The sensitive interfaces (`LocalSessions`, `CoworkSpaces`, `FileSystem`) share the gate's chunk and were already exempt on master. `IPC_ORIGIN_GUARD_RE` is anchored on the literal validation message, so a wider file set can't make it rewrite a non-origin guard.
- **`protocol-forwarder` candidate list** — no hijack. All three candidates live under the user-owned `$COWORK_DIR`, and `install_stubs` runs after `extract_archive`, so a file of that name inside a downloaded asar can't win the slot. The forwarder applies a length cap, `new URL()` parse, `claude:` protocol check and character allowlist before `execFile('bash', [launchSh, url])` — array form, no shell. Pre-PR that fast path was dead code, so the same URL reached `launch.sh` with *no* validation: enabling it adds validation.

Also cleared: `patch_host_platform` going markerless and global across ~300 files (all four `Unsupported platform` sites are platform gates — the project's stated purpose — and the fourth is unreachable with `platform` spoofed to `darwin`), and the `}`-bounded return-gate regex.

### One hardening note, not a finding

`install.sh:861` writes the launcher with an unquoted heredoc, and line 866 is `COWORK_DIR="${INSTALL_DIR}"`. Before this PR that interpolation was a constant; now it carries a user-settable value, so `$(…)` in `CLAUDE_INSTALL_DIR` would be written into `~/.local/bin/claude-desktop` and run on every launch. Not a vulnerability under the normal model — the user set their own variable — and **not something I want fixed in this PR.** I note it only because it's fine today and stops being fine the moment someone teaches the installer to read that value from a config file. Worth a follow-up to quote the heredoc.

Merging now, resolving the pinning-table conflict as described above. Thanks for a genuinely well-built PR — the verification discipline on it made this review much cheaper than it would otherwise have been.

---

chefboyrdave21:
Confirming the open question: **maintainer edits are enabled** on this branch (`maintainerCanModify: true`), so please push the pinning-table resolution straight to it whenever suits. Not rebasing, as asked.

On the security review — thank you for running it, and for writing up the `shellPathWorker` finding as a plant-and-execute path. Neither @shawnyeager nor I framed it that way; we both treated it as "PATH priming is broken." The bit I'd missed is that `resourcesPath` is redirected to a **user-writable** directory, so the missing file isn't just missing, it's plantable — and landing in the main process puts it on the wrong side of the very admission layer this project maintains. Useful correction to how I'll read "file resolves to a path that doesn't exist" in future.

Noted on `install.sh:861` — agreed it's not a finding today and shouldn't ride along here. Happy to send the heredoc quoting as a small follow-up PR once this lands, unless you'd rather fold it into whatever shape #170 takes, since both touch the generated launcher.

johnzfitch:
This is landing in #173 — apologies for the stall. It was approved and announced as merging on 2026-08-09, then the pinning-table resolution never actually got pushed, and it sat conflicting for a week through no fault of yours.

Merged with your 10 commits and authorship intact rather than squashed or re-authored. The conflict was exactly what was predicted on this thread: `COMPAT.md` only, both sides appending a pinning row, resolved by keeping both. You were asked not to rebase and you didn't — that was the right call, the resolution just needed someone to do it.

Verified after resolving:

```
tests/test-compat-pins.sh    11 passed, 0 failed
tests/test-cowork-patch.sh   58 passed, 0 failed
node --test tests/node/...  559 passed, 0 failed
```

58/58 matches what you and the review both reported independently, and `LAST_TESTED_ASAR_VERSION=1.26832.0` now resolves to your `.zip` URL and checksum.

This PR should close itself as merged once #173 lands, since its head commit becomes reachable from `master`. Leaving it open until then rather than closing it manually, so it closes as *merged* and the credit lands where it belongs.

Two things from this thread that #173 does **not** do, so they don't get lost:

- **#170** stays open and untouched. It's waiting on a call between your three refactor shapes, and guessing at that would waste the work you put into scoping it.
- The `install.sh:861` heredoc quoting you offered as a follow-up is still open and still worth doing.

One thing you may want to know: while chasing CI here, the AUR publish workflow turned out to have failed on **every run since 2026-04-23**, dying on a malformed signing key. That's why the AUR package still serves `1.1.4010-10` — your observation on this PR that the `pkgver` would come out `1.26832.0-1` was pointing straight at it. #173 makes that failure self-diagnosing, though the secret itself still needs re-adding by hand.

---

PR #171, opened 2026-08-13
fix: support Claude Desktop 1.28929

PonlawatP:
## What changed

- Discover and patch both `index.chunk-*` and `index2.chunk-*` main-process bundles in the installer, launcher, and PKGBUILD.
- Generalize the Cowork platform-gate matcher for `let`/`var` declarations and single, double, or template-literal quotes.
- Apply auxiliary IPC and platform patches even when the Cowork platform gate lives in a different chunk.
- Accept both `throw Error(...)` and `throw new Error(...)` IPC origin guards.
- Allow host `Read` to consume SDK-spooled MCP results under `.claude/projects/*/tool-results/*` while retaining the subsequent projects-root containment check.
- Record Claude Desktop `1.28929.0` as tested with its CDN URL and SHA-256.

## Why

Claude Desktop `1.28929.0` changed the Vite bundle layout and minified syntax. The Cowork gate moved to `index2.chunk-*`, while IPC guard sites remained distributed across other chunks. The existing patcher therefore reported that it could not find the platform gate and skipped auxiliary patches.

After those fixes, Visualize exposed a second Linux-specific failure: large MCP results are spooled to SDK-owned tool-result files, but the host safe-path resolver rejects the session storage prefix as an automount root before the normal allowed-root check runs. This prevented `visualize.read_me` from reaching `visualize.show_widget` reliably.

## Impact

Claude Desktop `1.28929.0` now installs and launches with Cowork enabled. Internal MCP Apps such as Visualize can read their spooled instructions and render their iframe widgets on Linux.

Verified on Fedora 44 with GNOME Wayland and Electron 43.2.0. The Visualize iframe reported both `firstrender` and `complete`.

## Checks

- `tests/test-cowork-patch.sh`: 43 passed, 0 failed
- `tests/test-compat-pins.sh`: 11 passed, 0 failed
- `node --test tests/node/current-path/*.test.cjs`: 559 passed, 0 failed
- `git diff --check`

`tests/test-install-paths.sh` still has its existing archive-harness failures: it checks the removed `fetch-dmg.py` and sends the current ZIP release through the DMG/7z extraction path. Those failures are outside this patch.

### DIFF
--- a/COMPAT.md
+++ b/COMPAT.md
@@ -5,8 +5,8 @@ to work with claude-cowork-linux. `install.sh` and the launcher grep the
 machine-readable lines below; the table further down is for humans.
 
 <!-- machine-readable; do not remove the next two lines -->
-<!-- LAST_TESTED_ASAR_VERSION=1.19367.0 -->
-<!-- LAST_TESTED_DATE=2026-07-15 -->
+<!-- LAST_TESTED_ASAR_VERSION=1.28929.0 -->
+<!-- LAST_TESTED_DATE=2026-08-14 -->
 
 ## Tested versions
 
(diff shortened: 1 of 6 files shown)


johnzfitch:
Recording the disposition of this one, since it closed quickly and part of it shouldn't be lost.

The patcher changes here overlap #167, which covered the same three 1.26832.0 breakages — the `index2.chunk-*` series, quote-agnostic gate matching for backticks and `let`, and auxiliary passes no longer keyed off the gate's own chunk. #167 went through a full review cycle and also carries the `discover_chunks()` glob, the `--doctor` split-build false negative, the `shellPathWorker` catch-all, and regression fixtures for the backtick shapes. It's landing in #173, so that part is covered.

Two things from here are carried forward rather than dropped:

**Your 1.28929.0 report is now in COMPAT.md**, as `[PARTIAL]` with the pinning row (URL + SHA-256) from this PR. It's attributed to this report rather than to the tree, because you verified it against your own patch set, not the one landing — the honest status is "expected to apply, since what landed is a superset, but nobody has run this tree against that build." If you do re-test on top of #173, that row can move to `[OK]` and take `LAST_TESTED_ASAR_VERSION` with it.

**The tool-result finding is #172.** That one has no equivalent anywhere in this tree and is the genuinely novel part of this PR: host `Read` can't reach SDK-spooled results under `.claude/projects/*/tool-results/*` because Linux's safe-path resolver rejects the prefix as an automount root before the allowed-root check runs, so MCP Apps like Visualize don't render. It isn't folded into #173 because `patch_tool_result_resolution()` substitutes raw `realpath` for the bundle's own safe-path resolver, which widens what host `Read` will open — #167's much smaller surface got an adversarial security review before landing, and this deserves the same rather than riding along in a merge cleanup. #172 records the mechanism, links your branch, and lists the two questions worth settling first. Your implementation is the starting point there, not a rejected one.

Worth saying: finding the Visualize failure at all meant getting past the patching breakage first and then noticing a second, unrelated problem underneath it. That's the part nobody else reported.

---

Issue #172, opened 2026-08-16
MCP Apps (Visualize) can't read SDK-spooled tool results on Linux — salvaged from #171


johnzfitch:
Splitting this out of #171 so it isn't lost when that PR closes as a duplicate.

#171 fixed the same three 1.26832.0 breakages as #167 (the `index2.chunk-*` series, backtick/`let` gate matching, auxiliary passes gated on the gate's own chunk). #167 is the more thorough of the two and has landed, so #171's patch logic is now redundant.

**One part of it is not redundant** and has no equivalent in this tree.

### The problem

Claude Code asks the host `Read` tool to consume SDK-owned results beneath:

```
<config>/projects/<session>/tool-results/<tool-id>.json
<local-agent-root>/<account>/<org>/local_<session>/outputs[/<artifact>]
```

Per @PonlawatP, the desktop's safe-path resolver on Linux rejects both storage prefixes as an "automount root" *before* the normal allowed-root check runs. Large MCP results get spooled to those files, so `visualize.read_me` can't reliably reach `visualize.show_widget` — internal MCP Apps don't render.

Reported on Fedora 44 + GNOME Wayland, Electron 43.2.0, asar 1.28929.0.

### Why it wasn't carried over

The proposed fix adds a `patch_tool_result_resolution()` pass to `enable-cowork.py` that rewrites `resolveFilePath(...)` call sites so paths matching an SDK-owned shape bypass the bundle's resolver and go through `require("node:fs/promises").realpath` instead.

That is security-sensitive in exactly the area this project is careful about. Substituting raw `realpath` for the app's own safe-path resolver widens what host `Read` will open, and the regex has to be right about which call sites it rewrites. #167's much smaller surface got a full adversarial review before landing; this deserves the same rather than riding along in a merge cleanup.

Two things I'd want established before it lands:

1. **What the resolver actually rejects, and why.** "Rejected as an automount root" is the reported symptom; the fix should be aimed at the specific check, and the narrowest patch that clears it.
2. **Containment of the bypass.** The `SDK_OWNED_RESULT_PATH_TEST` regex is anchored with `(?:^|\/)` and `$`, but `.claude/projects/.+/tool-results/[^/]+` uses a greedy `.+` that spans `/`. Worth confirming a crafted path can't match its way out of the session directory, and that the `.catch(() => { throw ... })` really does fail closed on every path.

### Prior art

[...]

Issue #170, opened 2026-08-09
PKGBUILD applies 3 of the 9 patch_index passes, so AUR installs miss the MCP/resourcesPath fixes


chefboyrdave21:
Split out of #167 at the maintainer's request — pre-existing and orthogonal to #166.

`launch.sh` runs **9** `patch_index` passes. `PKGBUILD`'s `build()` runs **3**. Because `/usr/bin/claude-cowork` execs electron directly against the packaged asar and never calls `launch.sh`, the missing six are never applied for anyone who installs from the AUR:

| # | Pass | `launch.sh` | `PKGBUILD` |
|---|------|:-----------:|:----------:|
| 1 | titlebar (main window) | yes | yes |
| 2 | titlebar (about window) | yes | yes |
| 3 | origin validation for `file://` preloads | yes | yes |
| 4 | `--effort xhigh` -> `max` | yes | **no** |
| 5 | Handoff `invalidateCurrentActivity` no-op | yes | **no** |
| 6 | Handoff `setUserActivity` no-op | yes | **no** |
| 7 | `resourcesPath` lookups -> asar-internal `resources/` | yes | **no** |
| 8 | MCP node-host asar paths -> `getAppPath()` | yes | **no** |
| 9 | unguarded `resourcesPath`+`app.asar` joins -> `getAppPath()` | yes | **no** |

`mainView.js`'s preload-origin patch is also launcher-only, and `install.sh` picks up 4–9 only because it invokes `launch.sh` at run time.

### Evidence

Measured on a package built from a branch that already carries **all** of #167's `PKGBUILD` fixes, so this is not "the recipe didn't get the fix" — the recipe simply never had these passes:

```
$ grep -ac 'process\.resourcesPath,`app\.asar`' /usr/lib/claude-cowork/app.asar
4
$ grep -ac 'process\.resourcesPath,`app\.asar`' ~/.local/share/claude-desktop/.asar-cache/app.asar
0

$ grep -ac 'xhigh"?"max"' /usr/lib/claude-cowork/app.asar
0
$ grep -ac 'xhigh"?"max"' ~/.local/share/claude-desktop/.asar-cache/app.asar
1
```

`/usr/bin/claude-cowork` ends in:

```sh
exec electron /usr/lib/claude-cowork/app.asar ${_sandbox_flag} --disable-gpu --class=Claude ...
```

No `launch.sh`, so there's no run-time pass to compensate.

### User-visible effect

[...]

PR #173, opened 2026-08-16
Land #167, fix the silently-broken AUR publish, and add test CI

### LINKED ISSUE #166: asar 1.26832.0: every patch silently no-ops (platform gate not found)

johnzfitch:
Fixes #166.

## What does this change?

Three things, in the order they were found.

**1. Lands #167.** It was reviewed, approved ("no further changes requested"), and announced as merging on 2026-08-09 — then never merged. When #169 landed the 1.19367.0 pinning row the branch went conflicting, and the resolution was never pushed. Merged here with the contributor's 10 commits and authorship intact. As predicted on the PR, only `COMPAT.md` conflicted: both sides append a pinning row, and the resolution keeps both. That closes #166.

**2. Fixes CI and packaging automation.**

`aur-publish.yml` has failed on **every run since 2026-04-23** — last success 2026-03-26, six consecutive failures. The cause is one line in the deploy action's container log:

```
Load key "/home/builder/.ssh/aur": invalid format
```

Nothing surfaced it, so the AUR package quietly stopped updating and still serves `1.1.4010-10`. **The `AUR_SSH_KEY` secret has to be re-added by a maintainer — this PR cannot fix that.** What it does is check the key up front, where the failure names itself and says what to do, instead of dying opaquely inside a container. It also runs the packaging tests before publishing (`test-cowork-patch.sh` exercises `PKGBUILD`'s own `patch_index` block, so a recipe that fails it shouldn't reach AUR), and bumps `actions/checkout` to v5 for the Node 20 deprecation already warning on every run.

`.SRCINFO` had drifted from `PKGBUILD`: `pkgrel 9` against the recipe's `12`, missing the `curl` and `zstd` depends, and listing `npm` where the recipe needs `asar`. Since AUR helpers resolve dependencies from `.SRCINFO`, that was wrong dependency resolution for AUR users, not just a stale version string.

And the repo ran **no tests automatically at all** — #167 sat with zero check runs. New `ci.yml` runs on push and PR:

| Job | What |
|---|---|
| `node` | `node --test tests/node/current-path/*.test.cjs` — 559 tests |
| `shell` | `test-compat-pins.sh` (11) + `test-cowork-patch.sh` (58) |
| `python` | byte-compile `enable-cowork.py` |
| `srcinfo` | regenerate `.SRCINFO` on Arch and diff, so this drift can't recur |

`test-install-paths.sh` is deliberately not gated: it wants a real archive from Anthropic's CDN and has pre-existing failures unrelated to any diff.

**3. Brings the compat reports current.** #171 duplicated #167's patch logic — same three 1.26832.0 breakages, less thoroughly — so its patcher changes are redundant now. Its 1.28929.0 report is not, and is recorded as `[PARTIAL]`, attributed to that report rather than to this tree, since it was verified against a different patch set. Its one genuinely unique fix (MCP Apps reading SDK-spooled tool results) is **not** folded in — it rewrites security-sensitive path resolution and deserves its own review, so it's captured in full detail in #172.

`LAST_TESTED_ASAR_VERSION` deliberately stays at **1.26832.0**: that's the newest build exercised end-to-end against this code, and it's the line the installer's "newer than tested" warning keys off. Moving it to a version nobody has run here would suppress a warning that's still earned.

Also corrects the test counts in README and CONTRIBUTING, which claimed "215+ tests across 18 files" in four places against an actual 559 across 35.

## Type

- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Refactor

## Testing

Full suite on the merged tree, matching what both @chefboyrdave21 and the maintainer reported independently on #167:

```
tests/test-compat-pins.sh    11 passed, 0 failed
tests/test-cowork-patch.sh   58 passed, 0 failed
node --test tests/node/...  559 passed, 0 failed
python -m py_compile enable-cowork.py   clean
```

**CI is green on this PR** — all four new jobs pass, which also exercises the workflow itself.

Two things I could not verify from the authoring container, called out rather than assumed:

- **No Docker daemon**, so `.SRCINFO` was written to `makepkg --printsrcinfo`'s field order by hand rather than generated. The new `srcinfo` job then regenerated it on a real Arch container and diffed — **it passes**, so the file is confirmed byte-identical to real `makepkg` output. (That job caught a genuine bug in itself first: `archlinux:base-devel` ships no `git`, so `actions/checkout` unpacks a tarball and a stray `git config` line failed. Fixed in the second commit.)
- **`ssh-keygen` isn't installed there**, so the AUR key precheck's behaviour is reasoned from `ssh-keygen -y -P ''` semantics, not executed. It fails closed with an actionable message, and `< /dev/null` keeps an encrypted key from prompting and hanging the job rather than failing it. First real exercise will be the next publish run — which cannot happen until the secret is fixed regardless.

- Distro / desktop: n/a — CI, packaging metadata, and docs; no runtime code in this diff beyond #167's own, which was verified on Manjaro + Cinnamon/X11, Electron 43.1.1, asar 1.26832.0
- Tested with `./install.sh --doctor`: not re-run here; #167 reported 20 passed / 1 warning / 0 failed
- Tested a full Cowork session: not re-run here; #167 reported sign-in and the Cowork surface working

## Security impact

Two notes, neither a new exposure:

The merged #167 changes process spawning and path resolution, and was cleared by an adversarial review on the PR with no exploitable findings. That review recorded the `shellPathWorker` catch-all as a **security fix**: `process.resourcesPath` is redirected to a user-writable directory, so the unguarded `app.asar` join resolved to a path anything writing under `~/.config` could plant and have executed in the Electron main process. This lands that fix.

Separately, the `patch_tool_result_resolution()` pass from #171 was **deliberately excluded** rather than carried along. It substitutes raw `realpath` for the bundle's own safe-path resolver on paths matching a regex, which widens what host `Read` will open. #172 records it with the two questions worth settling before it lands.

No credentials, tokens, or log output with secrets are in this diff. The key precheck reads `secrets.AUR_SSH_KEY` into a `600` file in the runner tmpdir, never prints it, and deletes it on both paths.

## Checklist

- [ ] `./install.sh --doctor` passes cleanly on my system — not run; no local install in this environment

## Needs a maintainer, not a merge

**The `AUR_SSH_KEY` secret is still broken.** Nothing in this PR can fix it — it needs re-adding by hand with the full unencrypted private key including its trailing newline. Until then AUR stays at `1.1.4010-10` and the publish job will now fail fast with that explanation instead of an opaque container error.

## Follow-ups this does not do

- **#170** (PKGBUILD runs 3 of 9 `patch_index` passes) is untouched. It's waiting on your call between the three refactor shapes @chefboyrdave21 sketched, and guessing at it would be the wrong move. The new `srcinfo` job is adjacent but doesn't address it.
- The `install.sh:861` unquoted-heredoc hardening, noted on #167 as "fine today, not fine once someone reads that value from a config file", is still open.


johnzfitch:
Both review comments were right. Fixed in 5706b38.

**`enable-cowork.py` failure hint** — correct, and it mattered more than it looks: since #167 the script runs across `index.js` *and* every `index*.chunk-*.js`, so a failure in `index2.chunk-CQIegP9t.js` was sending people to `index.js`, which is exactly the file where the gate *isn't* on the builds this PR fixes. Now prints the path it actually inspected.

While in there — the same line printed literal doubled braces:

```
a function checking process.platform and returning {{status:"unsupported"}}.
```

Brace-doubling is only an escape inside an f-string, and that one is a plain string. Switched to a single-quoted string so nothing needs escaping.

**`install.sh` non-empty install dir** — also correct. Worth recording that this one is **pre-existing on master**, not introduced here; `git clone` has always refused a non-empty destination, so "will overlay" was never true and the user got a bare clone error one step later. Fixed here anyway because this PR is what makes it reachable: #167 added `CLAUDE_INSTALL_DIR`, so the directory is now any path the user names rather than a fixed one under `~/.local/share`.

Now fails up front with what to do about it. Checked all four branches, since the guard has to agree with what `git clone` actually rejects:

| Install dir | Behaviour |
|---|---|
| absent | proceeds to clone |
| empty | proceeds to clone |
| non-empty | fails early with the explanation |
| dotfile-only | fails early — `git clone` rejects these too |

Suite still green: 11/11 pins, 58/58 patch, 559/559 node.

---

Issue #132, opened 2026-06-01
Latest supposedly tested version (1.6259.1) does not allow any session to start


elvanor:
### Area

Cowork session (spawn / spawn fails)

### What happened?

According to COMPAT.md, 1.6259.1 is tested and works. I installed this version but it fails to spawn a cowork session. The app itself starts, and chat conversations work. But the cowork stuff does not.

elvanor@caradhras ~/desktop/claude-cowork-linux $ CLAUDE_ARCHIVE=/home/elvanor/desktop/Claude-1.6259.1.dmg ./install.sh 

==========================================
 Claude Desktop for Linux - Installer
 Version: 4.0.0
==========================================

[INFO] Checking dependencies...
[OK] Found: git
[OK] Found: 7z
[OK] Found: node
[OK] Found: npm
[OK] Found: asar
[OK] Found: electron
[OK] Found: bwrap
[OK] Node.js version OK (v24)

[INFO] Updating existing installation...
Already up to date.
[OK] Repository updated

[INFO] Using user-provided archive: /home/elvanor/desktop/Claude-1.6259.1.dmg

[INFO]   File:    Claude-1.6259.1.dmg
[INFO]   Size:    303MB
[INFO]   Date:    2026-06-01 12:01:06
[INFO]   Latest:  v1.9659.2
[INFO]   SHA256:  d4bd47a08019de6f...

[INFO] Extracting archive...
[INFO] Extracting app.asar...
[OK] Extracted app to linux-app-extracted/

[INFO] Installing stubs from /home/elvanor/desktop/claude-cowork-linux...
[INFO] Syncing stubs and launch scripts to install dir...
[OK] Stubs installed

[INFO] Applying cowork patch...
SUCCESS: Patched /home/elvanor/.local/share/claude-desktop/linux-app-extracted/.vite/build/index.js
  rLi() now returns {status:"supported"} unconditionally
  getHostPlatform() patched: throw replaced with return"darwin-x64"
  IPC origin guards: no matching sites found
  Platform return gates patched: 1 sites neutralized
[OK] Patches applied

[OK] Created launchers: ~/.local/bin/claude-desktop, ~/.local/bin/claude-cowork

[OK] Environment configured

Installed sizes: [128, 256, 512, 1024]
[OK] App icon installed to hicolor theme

[OK] Version sentinel: 1.6259.1

[INFO] Running post-install checks...

==========================================
 Claude Desktop for Linux - Doctor
[...]

PR #175, opened 2026-08-16
fix(#170): one shared patch pass list, plus the silent-failure bugs it surfaced

### LINKED ISSUE #170: PKGBUILD applies 3 of the 9 patch_index passes, so AUR installs miss the MCP/resourcesPath fixes

johnzfitch:
## What does this change?

Closes #170, and fixes four more bugs of the same family found while tracing it. The theme throughout: **a step that did nothing, and said it worked.**

### 1. One shared patch pass list (#170)

`launch.sh` had grown to 9 `patch_index` passes while `PKGBUILD`'s `build()` still applied 3. `/usr/bin/claude-cowork` execs electron directly against the packaged asar and never calls `launch.sh`, so AUR users never got the other six — nor the `mainView.js` preload-origin patch, which was launcher-only for the same reason.

Following the steer on the issue, the passes move into a new `patch-index.sh` that both consumers source, so the drift is unrepresentable rather than merely tested for. `install.sh`'s private copy of the chunk discovery collapses onto the same file, taking it from three copies to one.

**Feasibility** (the crux the issue flagged): no `source=()` change needed. `PKGBUILD`'s only source is `git+https://github.com/johnzfitch/claude-cowork-linux.git`, so makepkg clones the whole repo to `${srcdir}/claude-cowork-linux`, and `build()` already reaches into it to run `${_repo}/enable-cowork.py`. A repo-root `patch-index.sh` arrives by exactly that mechanism.

**Impact is narrower than the issue implies** — @chefboyrdave21 confirmed this independently on their own install and owns the original mis-attribution:

| Pass | Effect on AUR |
|:--|:--|
| unguarded `resourcesPath`+`app.asar` join | **genuinely broken, now fixed** — the trampoline pins `process.resourcesPath` to `/usr/lib/claude-cowork/resources` while the asar sits at `/usr/lib/claude-cowork/app.asar`, so `shellPathWorker.js` resolved to a nonexistent file and login-shell PATH priming failed |
| `mainView.js` preload origin | **genuinely broken, now fixed** — its absence leaves CoworkSpaces unexposed to the renderer (empty Projects page, spaces don't persist) |
| `resourcesPath` lookups, MCP node-host | inert either way — both are `app.isPackaged`-gated, and both launchers start electron with a *path argument*, so `process.defaultApp` is true and `isPackaged` is false; those seds rewrite a branch that is never taken |
| Handoff ×2, `--effort xhigh` | already covered by `frame-fix-wrapper.js` and `session_orchestrator.js`, both of which `PKGBUILD` does install |

This is **not** the cause of "MCP Filesystem: Node host not found". The issue title still carries that claim; `patch-index.sh`'s header now says so explicitly so nobody re-adds the inert passes expecting a fix.

### 2. `enable-cowork.py` corrupted the bundle and reported SUCCESS

`HOST_PLATFORM_THROW_RE` matched the `Error(...)` argument with `[^)]*`, which stops at the first `)`. On a bundle whose message is built with a call in it —

```js
throw new Error("Unsupported platform: "+getPlatformName())
```

— the match ended *inside* that call, and the substitution left its closing paren behind as `return"darwin-x64")`. That is a syntax error. `patch_host_platform` still returned `True`, so the script printed `SUCCESS`, exited 0, and both callers counted the file as patched.

Reproduced end-to-end. Every existing fixture happens to use a paren-free message, which is why nothing caught it. Now matched paren-aware; deeper nesting **fails closed**. A post-patch `node --check` warns on stderr when the result is unparseable, deliberately **without** changing the exit code — both callers read non-zero as "no platform gate here", so failing that way would *hide* the corruption.

### 3. The `mainView.js` pass repacked the asar on every launch

`sed -i` rewrites its target whether or not the pattern matched, so a miss still bumps mtime — and that pass guarded only on the marker it inserts, which a miss also leaves absent. On any build where the preload's identifier isn't `e`, it re-ran every launch and made `launch.sh`'s "anything newer than the cached asar" check true forever. Tests pin **mtime**, not content — content was already unchanged, which is exactly why this stayed invisible.

### 4. `launch-devtools.sh` debugged the previous build

It synced the stubs and never repacked, so electron got whatever `launch.sh` last packed. For a debugging tool that's the worst shape of failure — indistinguishable from "my change was wrong". Relatedly `launch.sh` ignored `asar pack`'s exit status, falling through to launch against a stale asar with the error already scrolled off.

### 5. `tests/test-install-paths.sh` could not pass, and nothing ran it

Stage 1 `py_compile`d `fetch-dmg.py`, which became `fetch-dmg.js` — stage 2 is even *titled* "fetch-dmg.js output modes". `main()` exits 1 on any failure, so the whole harness exited non-zero regardless. Nothing noticed because the file is in no CI job. Single-arg stage selection also ran N-through-8 rather than "just stage N" as documented, so asking for one static stage tried to download a DMG.

Stage 1 is now wired into CI — pure static analysis, no network or Docker. That's the actual fix; the `fetch-dmg` drift was the symptom.

### 6. Comment corrections

`ipc_overrides.js` said keys match via `channel.includes(key)`; `matchOverride` uses `endsWith`. Not cosmetic — anyone auditing for collisions off the old comment reaches the wrong conclusion. `linux_ipc_stubs.js` named three insertion points in `frame-fix-wrapper.js` that no longer exist; that removal is worth recording, since a generic fallback answering unknown methods from a default table is exactly the shape that would forge an answer to a security question.

### 7. Fixes from the pre-merge review

An adversarial review of the combined `#175 + #176` tree caught four things, three of them regressions this PR itself introduced:

- **`install.sh` stopped syncing `stubs/`.** Refactoring the copy loop for the earlier review comment silently dropped `cp -rf "$stub_src/stubs" "$INSTALL_DIR/"`. Invisible at install time, because the stubs are already in the extracted tree — it bites on the *next* launch, when `launch.sh` re-syncs from the stale `$INSTALL_DIR/stubs` over the fresh ones.
- **Two `die()`s aborted a run master completes.** They land *after* the destructive `asar extract`, skipping `create_launcher`, the sentinel and the doctor. Deterministically reachable: a tag-pinned `CLAUDE_REPO_REF` leaves `$INSTALL_DIR` on a detached HEAD, `git pull --ff-only` then fails permanently, and `$INSTALL_DIR` never receives `patch-index.sh`. Both now warn loudly, name the repair command, and fall back to the same `find(1)` master used — never worse than master, and the doctor still catches an unpatched install.
- **`patch-index.sh` warned on every launch.** The launchers re-patch a *persistent* tree, so from the second launch on every pass legitimately matches nothing. Measured: 0 warnings fresh, **9 on every run after** — noise, and it made a real #166 silent-no-op indistinguishable from the steady state. Off in the launchers; `PKGBUILD` keeps it, since `build()` always starts from a fresh extract where a miss is real signal.
- **`python` was a build dependency listed only in `optdepends`.** `build()` runs it unconditionally, so a clean-chroot build had no python and failed in the patch loop. Moved to `makedepends`; verified the in-array comment leaves the array as exactly `(p7zip asar curl python)`.

## Type

- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [x] Refactor

## Testing

```
tests/test-cowork-patch.sh    83 passed, 0 failed   (was 58)
tests/test-compat-pins.sh     11 passed, 0 failed
tests/test-install-paths.sh 1  6 passed, 0 failed   (was a guaranteed fail)
node --test .../current-path  559 passed, 0 failed
```

Also verified against the **merged** `master + #176 + #175` tree: clean merge, all suites green, the generated launcher `bash -n`s and keeps a hostile `INSTALL_DIR` literal.

Every fix was reproduced against the pre-fix code first, and each regression test was checked to actually fail without its fix.

- Distro / desktop: not run against a live desktop — these are CI-verifiable changes; the suites exercise the passes against fixtures in both double-quoted and 1.26832.0 backtick shapes
- Tested with `./install.sh --doctor`: not runnable here (no Claude Desktop install in this environment)
- Tested a full Cowork session: no — **an AUR build on a real Arch box remains the check worth doing**, since `build()` is the path that changed most

## Security impact

- [x] This change does not touch credential handling, token passthrough, or process spawning
- [ ] This change does touch security-sensitive code — explanation below:

The patch passes move verbatim — same guard patterns, same sed scripts, same order, including the pass-8-before-pass-9 ordering constraint and the do-not-simplify-the-disclaimer-wrapper note (#132). No pattern was reformatted or re-quoted; the character-for-character `["`]` alternation is what survived #166. The `enable-cowork.py` regex change is strictly *narrowing* in the failure direction.

## Checklist

- [ ] `./install.sh --doctor` passes cleanly on my system — not runnable here, see Testing

### DIFF
--- a/.SRCINFO
+++ b/.SRCINFO
@@ -1,7 +1,7 @@
 pkgbase = claude-cowork-linux
 	pkgdesc = Anthropic Claude Desktop with Cowork (local agent) support for Linux
 	pkgver = 1.1.4010
-	pkgrel = 12
+	pkgrel = 13
 	url = https://github.com/johnzfitch/claude-cowork-linux
 	arch = x86_64
 	license = custom:proprietary
--- a/.SRCINFO
+++ b/.SRCINFO
@@ -8,6 +8,7 @@ pkgbase = claude-cowork-linux
 	makedepends = p7zip
 	makedepends = asar
 	makedepends = curl
+	makedepends = python
 	depends = electron
 	depends = nodejs
 	depends = curl
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,11 +46,19 @@ jobs:
       - name: Compat pin consistency
         run: bash tests/test-compat-pins.sh
 
-      # Extracts the patch_index blocks from launch.sh and PKGBUILD and runs
-      # them against fixtures, including the backtick/let/index2 shapes from
-      # #166 and the minified-name fallback.
+      # Exercises the shared patch-index.sh pass list against fixtures, including
+      # the backtick/let/index2 shapes from #166 and the minified-name fallback,
+      # and guards that launch.sh and PKGBUILD still both source that one list.
       - name: Cowork patch passes
         run: bash tests/test-cowork-patch.sh
+
+      # Stage 1 of the install harness is pure static analysis -- no network, no
+      # Docker, no Arch -- so it can run here even though stages 2-8 cannot.
+      # Nothing ran this file at all, which is how it came to py_compile
+      # fetch-dmg.py for however many months after that script became
+      # fetch-dmg.js. A harness nobody runs rots into a harness nobody can run.
+      - name: Install harness static analysis
+        run: bash tests/test-install-paths.sh 1
 
   python:
     name: enable-cowork.py syntax
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -1,7 +1,7 @@
 # Maintainer: Zack Fitch <zack@internetuniverse.org>
 pkgname=claude-cowork-linux
 pkgver=1.1.4010
-pkgrel=12
+pkgrel=13
 pkgdesc="Anthropic Claude Desktop with Cowork (local agent) support for Linux"
 arch=('x86_64')
 url="https://github.com/johnzfitch/claude-cowork-linux"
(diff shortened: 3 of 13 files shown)


chefboyrdave21:
Reviewed, and ran the suite on this branch: **77 passed, 0 failed.**

**First, the correction is yours to make and mine to own.** #170 asserted AUR users hit "MCP Filesystem: Node host not found." That was wrong, and your impact table is right: both launchers exec electron with a *path argument*, so `process.defaultApp` is true, `isPackaged` is false, and the MCP node-host and `resourcesPath` seds rewrite a branch that is never taken. I checked it here — `exec electron /usr/lib/claude-cowork/app.asar`, and zero "Node host not found" in the logs of either install.

I reasoned from "the pass is missing" to "the symptom must occur" without checking the branch was reachable. The measurements in the issue were real (4 unpatched joins, no `xhigh` remap), but I attached the wrong user-visible consequence to them. Of the six missing passes, two mattered — the unguarded join, and the `mainView.js` preload origin, which I hadn't flagged at all and which is arguably the worse of the two.

### One finding: `patch-index.sh` inherits the claim you corrected

The new file's header carries my wrong version forward, at `patch-index.sh:24`:

```
# six -- most visibly the MCP node-host path fix, which surfaces as
# "MCP Filesystem: Node host not found" and no local MCP servers.
```

That's the opposite of this PR's own impact table, which has that pass as inert either way. The same file even states the reason 140 lines later, at `:166`:

```
# when app.isPackaged is false (which it always is when running via `electron .asar`)
```

So the file contains both the correct fact and the wrong conclusion drawn from its absence. Given item 6 of this PR is about comments that make an auditor reach the wrong conclusion, this one seems worth fixing in the same pass — it's the header of the file, so it's the first thing anyone touching the pass list reads, and it would send them chasing an MCP bug that isn't there.

Suggested: name the two passes that actually regressed on the AUR path (unguarded `resourcesPath`+`app.asar` join -> broken `shellPathWorker.js`; missing `mainView.js` preload-origin patch -> CoworkSpaces unexposed, empty Projects page), and note the isPackaged-gated pair are inert under both launchers so nobody re-adds them expecting a fix.

### On the rest

The `[^)]*` corruption is the one I'd have least liked to ship — `SUCCESS` + exit 0 on output that doesn't parse is the same silent-success shape as #166, one layer down. Failing closed on deeper nesting is the right call over best-effort rewriting.

`launch-devtools.sh` debugging the previous build is a nasty one; "my change had no effect" is indistinguishable from "my change was wrong," and you'd burn a lot of time before suspecting the tool. Same for `asar pack`'s ignored exit status — launching against a stale asar while the error scrolls past is exactly the failure mode this project keeps producing.

Nothing else from me. Happy to send the header-comment fix as a commit here if you'd rather not context-switch.

PR #177, opened 2026-08-24
fix(welcome): strike Mac, append Linux on the Cowork gate welcome screen

### LINKED ISSUE #174: Welcome screen greets Linux users with "Claude for Mac" (the darwin spoof leaks into the UI)
Signed out, the welcome screen on Linux reads **Claude *for* Mac**.

That's not the webapp being wrong — it's us. The heading picks its platform word from the app's own mac/linux/windows branch, and we spoof `process.platform` to `darwin` so the Cowork gate returns `supported`. The spoof does its job and then leaks one string into the UI. Un-spoofing isn't an option; that's the whole mechanism.

So it's a cosmetic artifact of the compatibility layer rather than an upstream bug, which is why I'm raising it here rather than letting it sit as folklore.

### What I'd propose rendering instead

> Claude *for* M̶a̶c̶ Linux

I want to make the case that the strikethrough is the *accurate* option, not just the cute one. Plain "Claude for Linux" would imply an official Linux build exists — which is precisely the thing this project is careful not to imply. The struck version states what's actually running: the Mac build, on Linux. It's self-documenting about what the repo does, and it costs one line.

(It also means the first thing a new user sees is a small, honest admission of what they've just installed, which felt more in keeping with the project than pretending.)

### How, and the part I'd want you to push back on

This has to be a DOM rewrite, not a bundle patch. I tried the bundle first and it doesn't work: the renderer is served live from `assets-proxy.anthropic.com`, and the bundled `resources/ion-dist` copy is a fallback this screen never loads. I confirmed that the hard way — planted a marker string in the local bundle, verified it reached the packed asar the running Electron had open, and watched it never render.

So the working version rewrites the rendered text from the existing `web-contents-created` hook in `frame-fix-wrapper.js`: walk text nodes, find one that is exactly `Mac` under a heading whose text is `Claude for Mac`, replace with `M̶a̶c̶ Linux`. U+0336 combining overlays keep it a plain string, so no JSX factory alias is involved and it survives minifier rotation.

**The mechanism is the bit I'd expect you to weigh, not the string.** Everything else in the repo patches a local bundle we control. This injects script into a renderer showing Anthropic's live web app, which is a new category for the project, and "we mutate their DOM" is a precedent you may simply not want — declining on that basis is completely reasonable and I won't re-litigate it.

Three limitations, so they're not discoveries later:

- **English-only.** The match needs literally `Claude for Mac`. Product names get localised elsewhere in the i18n resources (`"Abrir Claude para escritorio"`), so other locales almost certainly no-op. Correct failure mode, but a partial fix.
- **Per-navigation cost.** It runs on `dom-ready` and `did-navigate` and arms a `MutationObserver` on the document subtree. Small, but non-zero, for one string.
- **Coupled to someone else's markup.** If the heading's shape changes it silently stops working. It degrades to a no-op rather than breaking the page, which is the right direction, but it's still their markup.

### If you want it without the injection

A plain-text `Claude for Linux` in the bundled `ion-dist` fallback is *not* equivalent — that copy isn't what renders — so there's no cheap non-injection version of this. The honest alternatives are: take the DOM rewrite, gate it behind a pref for people who want it, or close this as working-as-intended and leave "Claude for Mac" as a known cosmetic wart worth a line in the README.

I've been running the DOM version locally on 1.26832.0 since 2026-08-07 with no side effects, and I'm happy to send it as a PR, put it behind a flag, or drop it entirely — genuinely no attachment to the outcome. Happy to post before/after screenshots if the idea has legs; not cluttering the thread with them up front.

chefboyrdave21:
Closes #174.

Signed out, the welcome heading reads "Claude for Mac" on Linux because it picks its platform word from the darwin spoof we present to the Cowork gate. Un-spoofing isn't an option since that's the whole mechanism, so this is a DOM rewrite rather than a bundle patch: the renderer for this screen is served live from `assets-proxy.anthropic.com`, and the bundled `resources/ion-dist` fallback never loads it (confirmed by planting a marker string there and watching it never render).

`buildMacLinuxRewriteScript()` walks text nodes for one that is exactly `Mac` under an ancestor (within 5 levels) whose trimmed text is `Claude for Mac`, and replaces it with `M̶a̶c̶ Linux` using U+0336 combining overlays. Wired into the existing Linux-only `dom-ready`/`did-navigate` handlers in `frame-fix-wrapper.js` (same block that already inserts the drag-region CSS fix), with a `MutationObserver` to reapply across SPA remounts.

Known limitations (also called out in the issue):
- English-only: matches the literal string `Claude for Mac`; other locales no-op.
- Small non-zero per-navigation cost (one DOM walk + an observer).
- Coupled to Anthropic's markup; degrades to a no-op (not a break) if the heading's shape changes.

Verification: `node -c` on the edited file, plus a standalone check that the generated script is valid JS, contains the match string, and that the strike-overlay helper produces `M̶a̶c̶` for `"Mac"`. No existing test in `tests/` covers this file's DOM-injection paths.

### DIFF
--- a/stubs/frame-fix/frame-fix-wrapper.js
+++ b/stubs/frame-fix/frame-fix-wrapper.js
@@ -602,6 +602,45 @@ function coworkSpaceRoute(targetUrl) {
   }
 }
 
+// Issue #174: the welcome screen's heading picks its platform word from the
+// darwin spoof we present to the Cowork gate, so signed-out Linux users see
+// "Claude for Mac". Un-spoofing isn't an option -- that's the whole
+// mechanism -- so this rewrites the rendered text node in place: strike
+// through "Mac" (U+0336 combining overlay, so it's a plain string survives
+// minifier rotation) and append "Linux". A MutationObserver reapplies it
+// across SPA navigation since the heading can remount.
+function buildMacLinuxRewriteScript() {
+  return [
+    '(function(){',
+    '  try {',
+    '    var STRIKE = "\\u0336";',
+    '    function struck(s) { return s.split("").join(STRIKE) + STRIKE; }',
+    '    function rewrite() {',
+    '      var walker = document.createTreeWalker(document.body, NodeFilter.SHOW_TEXT, null);',
+    '      var node;',
+    '      while ((node = walker.nextNode())) {',
+    '        if (node.nodeValue !== "Mac") continue;',
+    '        var el = node.parentElement, depth = 0;',
+    '        while (el && depth < 5) {',
+    '          if (el.textContent && el.textContent.trim() === "Claude for Mac") {',
+    '            node.nodeValue = struck("Mac") + " Linux";',
+    '            break;',
+    '          }',
+    '          el = el.parentElement;',
+    '          depth++;',
+    '        }',
+    '      }',
+    '    }',
+    '    rewrite();',
+    '    if (!window.__coworkMacLinuxObserver) {',
+    '      window.__coworkMacLinuxObserver = new MutationObserver(rewrite);',
+    '      window.__coworkMacLinuxObserver.observe(document.body, { childList: true, subtree: true, characterData: true });',
+    '    }',
+    '  } catch (e) {}',
+    '})();',
+  ].join('\n');
+}
+
 // Build the JS injected into the renderer to perform client-side navigation.
 // The route is JSON-encoded (never string-concatenated raw) so a hostile space
 // name/URL cannot break out of the string literal into executable code.
@@ -1428,9 +1467,11 @@ Module.prototype.require = function(id) {
         _appRef.on('web-contents-created', (_ev, contents) => {
           contents.on('dom-ready', () => {
             contents.insertCSS(_fixCSS).catch(() => {});
+            contents.executeJavaScript(buildMacLinuxRewriteScript()).catch(() => {});
           });
           contents.on('did-navigate', () => {
             contents.insertCSS(_fixCSS).catch(() => {});
+            contents.executeJavaScript(buildMacLinuxRewriteScript()).catch(() => {});
           });
           // Issue #147: convert a full-page navigation to a locally-created
           // Cowork space into an in-app SPA route change, so it loads from IPC


johnzfitch:
Adopted in #183, which carries your commit unchanged with authorship intact, plus one follow-up commit.

The rewrite itself is right and I didn't touch it — the tests I added confirm the same output against both revisions: `Claude for Mac` becomes `M̶a̶c̶ Linux`, and a bare `Mac` with no such ancestor is left alone. Finding that the screen is served live from `assets-proxy.anthropic.com` (and proving the bundled `ion-dist` fallback never loads it, with a planted marker) was the hard part, and it's what made a DOM rewrite the right shape instead of a bundle patch.

What the follow-up changes is cost, not correctness. The observer took `{ childList, subtree, characterData }` on `document.body` and ran a full-document `TreeWalker` in its callback. Those two together are the problem: `characterData` + `subtree` makes every streamed token a mutation record, and every record walked every text node in the document — quadratic in conversation length, for the life of the window. You flagged "small non-zero per-navigation cost" as a known limitation; I think it's larger than that reads, and it never disconnected, so a signed-in session kept paying it for a heading that had already stopped existing.

Bounded four ways in #183: disconnect on a successful rewrite, a 15s deadline for pages where the heading never appears, `childList` only (a remount inserts the heading as a subtree, which is a `childList` record), and a burst coalesced into one walk per frame. The tradeoff is now the other way round — a welcome screen that remounts mid-page with no navigation and after the deadline keeps its "Mac". A label, not a break.

Also worth noting: the early `return` when `document.body` is null. The original relied on the outer `try/catch`, which meant a `dom-ready` on a document with no body yet threw *past* the observer installation and left that page unwatched entirely.

Leaving this open so you can look at #183 first — close it whenever you're happy, or say the word if you'd rather take the observer fix into this branch yourself and land it here instead. Genuinely good catch on a bug most people would have shrugged at.

---

johnzfitch:
Closing as superseded: the adoption in #183 merged to master on 2026-08-30. Master carries your commit unchanged with authorship intact (bd23c03) plus the bounded-observer follow-up (a0d47f0, 890a878), and this branch now conflicts with master with nothing left in it that is not already landed. Thanks again for finding this one.

---

Issue #85, opened 2026-04-20
Release on flathub


spazziale:
### Category

New distro support

### What problem does this solve?

flathub release is universal and very convinient

### Proposed solution

release a flatpak

### Alternatives considered

using the installer

### Contribution

- [ ] I am willing to submit a PR for this

johnzfitch:
Triage notes on what a Flathub release would actually take, for anyone who wants to pick this up:

1. **No redistribution:** this project never ships Anthropic's binaries, so the manifest would need Flatpak's `extra-data` mechanism to download the Claude archive client-side at install time (the same model proprietary apps like Chrome use), plus the full extract/stub/patch pipeline running inside the flatpak sandbox.
2. **Self-mutating app tree:** `launch.sh` applies sed patches and repacks the asar in its own directory on every launch. A flatpak's `/app` is read-only, so the whole tree would have to be mirrored into `$XDG_DATA_HOME` at first run — the Nix flake just merged in #135 solves the identical problem with a writable per-user mirror, and would be the template to follow.
3. **Sandbox-in-sandbox:** Cowork uses bubblewrap, which generally cannot run nested inside the flatpak sandbox; this likely needs `flatpak-spawn --host` or sub-sandbox portals, and is the riskiest unknown.
4. **Base:** `org.electronjs.Electron2.BaseApp` + an Electron 41-compatible runtime, plus a Flathub maintainer willing to own update cadence (the asar version pinning in COMPAT.md would drive releases).

This is a meaningful chunk of work that won't make the current round — marking as help-wanted. If someone wants to attempt it, starting from `nix/package.nix` + `nix/claude-desktop-launcher.sh` as the architectural reference is the recommended path, and we're happy to review a draft manifest.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #178, opened 2026-08-24
draft(flatpak): starting-point manifest for #85

chefboyrdave21:
Draft for #85, opened as a draft on purpose: this has never been run through `flatpak-builder`, that wasn't available in the environment this was written in. Opening it because the triage on #85 explicitly invited a draft manifest for review, not because it's close to mergeable.

**What it does:** ports `nix/package.nix`'s build phase (extract → stub → `enable-cowork.py`) and `nix/claude-desktop-launcher.sh`'s writable-tree mirror almost verbatim, since `/app` is read-only the same way the Nix store is. Kept close to the Nix version on purpose so the two packaging paths don't drift into separately-maintained pipelines.

**What it doesn't do, and isn't pretending to:** solve the sandbox-in-sandbox problem you flagged as the riskiest unknown. Cowork spawns its own `bwrap`; nesting that inside Flatpak's sandbox is unresolved here. `--talk-name=org.freedesktop.Flatpak` is requested so a future `flatpak-spawn --host` fix to the sandbox spawner (the pattern distrobox/devcontainers-in-flatpak use) doesn't also need a manifest change, but no code change to `launch.sh` or the cowork sandbox spawner is attempted in this PR.

Also unverified/placeholder, listed in full in `flatpak/README.md`:
- `extra-data` sha256/size are placeholders, need `node fetch-dmg.js --json`.
- `org.electronjs.Electron2.BaseApp` runtime version (`24.08`) isn't checked against Electron 41 compatibility.
- `--filesystem=home` for attached folders is the blunt version; portal-based access is the right long-term answer.

Verification: YAML, the metainfo XML, the wrapper script, and each embedded `build-commands` shell block all pass a syntax check. No actual `flatpak-builder --user --install` run, no launch, no Cowork task attempted, no GPG/hardware here to do that with.

If anyone can run `flatpak-builder` against this, the sandbox step is almost certainly where it breaks first and is the most useful thing to report back on #85.

### DIFF
--- a/flatpak/README.md
+++ b/flatpak/README.md
@@ -1,0 +1,58 @@
+# Flathub draft (#85)
+
+**Status: draft skeleton, never built.** `flatpak-builder` was not available in
+the environment this was written in, so nothing here has been run — not
+`flatpak-builder --user --install`, not a launch, not a Cowork task. Read this
+before trying it, and please report back what breaks.
+
+## What this reproduces
+
+`io.github.johnzfitch.ClaudeCoworkLinux.yml`'s build-commands are a
+near-verbatim port of `nix/package.nix`'s `buildPhase`: extract Claude's
+archive, drop in the stubs, run `enable-cowork.py` once at build time. The
+runtime wrapper (`claude-desktop-wrapper.sh`) mirrors
+`nix/claude-desktop-launcher.sh`'s trick for the same reason: `/app` is
+read-only and `launch.sh` mutates its own directory (sed patches, `asar pack`
+into `.asar-cache/`) on every run, so the prepared tree gets mirrored into a
+writable per-user copy on first run or version bump. No extra permission is
+needed for that mirror because Flatpak already gives every app its own
+private `XDG_DATA_HOME`.
+
+## What's a guess, not a fact
+
+- **`extra-data` sha256/size are placeholders.** Run
+  `node fetch-dmg.js --json` and fill them in before attempting a build — an
+  extra-data source with a wrong hash just fails at install time, so this
+  isn't a silent risk, but it will stop you immediately if skipped.
+- **The runtime/base versions (`24.08`) are unverified against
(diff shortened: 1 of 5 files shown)


johnzfitch:
Closing this as a draft that has done its job — which was mapping the problem, not solving it. Your own README is the reason: `flatpak-builder` was never available, so nothing here has been built, installed, or launched once. Keeping it open as a draft implies it's a few commits from ready, and the thing standing in the way isn't commits.

**The blocker you named is the real one.** Cowork spawns `bwrap` itself to isolate task execution, and nesting that inside Flatpak's sandbox is unsolved here. Requesting `--talk-name=org.freedesktop.Flatpak` so a future `flatpak-spawn --host` approach doesn't need a manifest change is good forward thinking, but as you say, no code does that — and it's a change to the sandbox spawn path, testable only from inside a real Flatpak sandbox. That's the work #85 actually needs, and it doesn't depend on this manifest existing.

Two things worth carrying forward when someone picks it up:

**The build-commands have a bug that just got fixed everywhere else.** This line:

```
cp stubs/@ant/claude-native/index.js "$tree/node_modules/@ant/claude-native/index.js"
```

`index.js` require()s `./safe_fs.js` unconditionally at module load, so a tree with `index.js` alone throws `MODULE_NOT_FOUND` the moment the stub is required. The AUR and Nix recipes both had exactly this, shipping a stub that could not load; #182 fixed all four paths and added a check that greps the tree for any file copying `claude-native/index.js` by name. Whoever revives this should use `cp stubs/@ant/claude-native/*.js`, and CI will now say so if they don't.

**Mirroring `nix/package.nix` was the right call** and worth keeping. Two packaging paths that drift into two sources of truth is #170, and the value of this draft is largely that it didn't reinvent the pipeline.

Reopen or open a fresh PR whenever you have a build that runs — the manifest is a fine starting point, it just needs someone who can execute `flatpak-builder` at least once. Thanks for being straight about what was and wasn't verified; that's what made it easy to triage.

---

Issue #164, opened 2026-08-01
asar 1.24012.9: local MCP servers blocked by disclaimer-wrapper exec-capability check


mnstahlberg:
## Summary

Since asar version **1.24012.9** (fetched 2026-08-01 via `install.sh --force` / `claude-desktop --update`), local MCP servers configured in `claude_desktop_config.json` fail to connect. The UI shows "Server disconnected" immediately after the connection is opened, before any JSON-RPC traffic is exchanged.

## Root cause

Verified with `strace -f` on the Electron main process: the MCP server's binary is **never actually `execve()`'d** at the OS level. The failure happens entirely in JS, before spawn is ever attempted.

`startup.log` shows the exact rejection:

```
[exec-capability] disclaimer unwrap REJECTED user-mcp path: /home/user/.local/bin/desktop-commander
```

This build apparently routes local MCP server launches through the same macOS-native "disclaimer" consent wrapper (`Helpers/disclaimer <cmd> <args>`) that was previously only used for `git` operations (see the comment/history around `stubs/frame-fix/frame-fix-wrapper.js` and `stubs/cowork/exec_capability_registry.js`). `resolveDisclaimerCommand()` intentionally rejects the `user-mcp` capability category — by design, per the existing comment, to avoid letting an IPC-influenced `args[0]` execute arbitrary user-installed binaries. That's the right call for an *arbitrary* path, but it now also blocks **every legitimately user-configured MCP server**, since apparently all local MCP servers get funneled through this wrapper in this build.

A second, smaller bug compounds this: `resolve()`'s `USER_MCP_PREFIXES` only lists `bin/`-style directories (`~/.local/bin/`, `~/.npm-global/bin/`, etc.), not the directories those `bin` symlinks actually point into. Since `fs.realpathSync()` follows the npm-global symlink (e.g. `~/.local/bin/desktop-commander` → `~/.local/lib/node_modules/@wonderwhy-er/desktop-commander/dist/index.js`), any **npm-globally-installed** MCP server resolves to a path outside every whitelisted prefix and gets classified as fully unresolvable (`[exec-capability] BLOCKED: ...`) even before reaching the disclaimer check.

## Fix applied locally (verified working)

Patch to `stubs/cowork/exec_capability_registry.js`:

[...]

PR #181, opened 2026-08-28
fix(cowork): handle -- separator in disclaimer wrapper args

PonlawatP:
## Problem
Every Code-tab / local-agent-mode session on Linux fails immediately with
"Claude Code process exited with code 127", surfaced as a "Claude Code
crashed" dialog on every message sent.

## Root cause
`stubs/cowork/exec_capability_registry.js`'s `resolveDisclaimerCommand()`
read `args[0]` as the command unconditionally. HostCLIRunner (Code tab)
invokes the disclaimer wrapper as `disclaimer -- <cmd> <args...>` -- a
leading `--` separator, no command in `args[0]`. `cmd` became the literal
string `"--"`, which matches neither the `claude.app` regex nor the
basename check, so the unwrap returned `null` and the real disclaimer stub
(`exit 127`) ran -- same failure mode as #132, a different argv shape at a
different call site.

## Fix
Skip a leading `--` before reading `cmd`/`rest` in `resolveDisclaimerCommand`.
No change to admission logic -- `resolveClaudeCli()`/`resolve()` still decide
what's actually substituted; this only changes which array index is read as
the command.

## Scope / not touched
- No change to `resolve()`, the allowlist, or any capability class.
- Confirmed this is not a regression carried by an in-flight branch:
  `git diff master -- stubs/cowork/exec_capability_registry.js` is empty
  against this branch's parent history, so the bug predates and is
  independent of any other work in flight.

## Checks
- Added two regression tests in
  `tests/node/current-path/exec_capability_registry.test.cjs`: the `--`
  case, and a bare `--` with no following command.
- Confirmed the new test fails against the pre-fix code and passes after
  the fix.
- `node --test tests/node/current-path/exec_capability_registry.test.cjs`:
  42/42 passing.

## Security note
`exec_capability_registry.js` is the disclaimer/admission chokepoint that's
been through several dedicated security-review passes in this repo's
history (f41417e, 0e1deeb, ea31eda, 04d7b16), and I understand a prior PR
touching this area is still under audit -- flagging explicitly even though
this file isn't on CONTRIBUTING.md's named security-sensitive list. This
change only shifts which argv index is read as `cmd` when the first
element is a bare `--`; the actual admission decision is untouched and
still routes through `resolveClaudeCli()`/`resolve()`. It adds no new
command classes and does not touch the allowlist.

### DIFF
--- a/stubs/cowork/exec_capability_registry.js
+++ b/stubs/cowork/exec_capability_registry.js
@@ -246,8 +246,16 @@ function createExecCapabilityRegistry({
   // agrees with resolve() for every class is what keeps that true.
   function resolveDisclaimerCommand(args) {
     if (!Array.isArray(args) || args.length === 0) return null;
-    var cmd = args[0];
-    var rest = args.slice(1);
+    // Some call sites (e.g. HostCLIRunner/Code tab) invoke the wrapper as
+    // `disclaimer -- <cmd> <args...>`, a leading `--` separator with no
+    // command in args[0]. Skip it before reading the command, or cmd becomes
+    // the literal string "--", which matches nothing below and falls through
+    // to the exit-127 stub -- the same failure mode as #132 but for this
+    // callsite's argv shape instead of the path shape.
+    var offset = args[0] === '--' ? 1 : 0;
+    var cmd = args[offset];
+    var rest = args.slice(offset + 1);
+    if (cmd === undefined) return null;
     // The asar invokes the Claude CLI through the disclaimer wrapper using
     // whatever path it chose -- a macOS-style claude.app/.../Claude path, or
     // the SDK path it installed (claude-code-vm/<ver>/claude), or a native
--- a/tests/node/current-path/exec_capability_registry.test.cjs
+++ b/tests/node/current-path/exec_capability_registry.test.cjs
@@ -222,6 +222,32 @@ describe('exec_capability_registry', () => {
         assert.deepStrictEqual(result.rest, ['-p', 'hi']);
       });
 
+      it('resolves a claude path passed with a leading -- separator (Code tab / HostCLIRunner)', () => {
+        // HostCLIRunner (Code tab) invokes the wrapper as
+        // `disclaimer -- <cmd> <args...>` -- a leading `--` with no command
+        // in args[0]. Before this fix, cmd read args[0] unconditionally, saw
+        // the literal string "--", matched neither the claude.app regex nor
+        // the basename check, and fell through to the exit-127 stub on every
+        // Code-tab session -- same failure mode as #132, a different argv
+        // shape at a different call site.
+        const reg = createExecCapabilityRegistry({
+          homedir: os.homedir(),
+          resolveClaudeBinaryPath: () => '/usr/local/bin/claude',
+        });
+        const result = reg.resolveDisclaimerCommand([
+          '--',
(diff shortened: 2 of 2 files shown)


johnzfitch:
Thanks for this, and for flagging the security-sensitivity yourself — that's the right instinct for this file.

Closing it, because I traced the argv shape end to end against the bundle in this tree and could not find a call site that produces it. Laying out what I checked, so you can point at whichever step is wrong if I've missed something.

**There is one disclaimer wrapper, and it never emits a leading `--`:**

```js
function Ql(t){return process.platform!=="darwin"?t:{cmd:rpt(),args:[t.cmd,...t.args]}}
```

`args[0]` is `t.cmd` by construction. All 12 call sites go through it.

**HostCLIRunner reaches it the same way.** `Ezt(t, e)` calls `yi(n, t, …)`, and `yi(t, e=[], r={})` does `Ql({cmd: t, args: e})` — so `args[0]` is the resolved Claude binary path, and the runner's own args follow it.

**The SDK doesn't insert one either.** The spawn builds `J = mvt(c) ? [...a, ...z] : [...a, c, ...z]`, where `c` is `pathToClaudeCodeExecutable` and `a` is `executableArgs`. `mvt` is `t => ['.js','.mjs','.tsx','.ts','.jsx'].some(e => t.endsWith(e))`, which is true for the disclaimer binary, so the native branch is taken: `[...executableArgs, ...flags]`, i.e. `[claudePath, …]`.

**The only `"--"` array element in the whole bundle** is inside `git diff` argv (`["diff","--no-ext-diff","-M","--no-color",t,"--",...]`), and there `args[0]` is still the git path.

Two other things worth saying:

- The second test (`['--']` → `null`) passes unchanged against `master`. `basename('--') !== 'claude'`, so it falls through to `resolve('--', [])`, which already returns null. I ran it to be sure. So it documents existing behaviour rather than guarding the change.
- The change itself is harmless — it shifts which index is read when `args[0]` is a bare `--`, and admission still routes through `resolveClaudeCli()`/`resolve()`. My objection isn't risk, it's that this file's comments are load-bearing (they're what stopped #132 and #164 recurring), and adding one that explains a call site nobody can find makes the next person's reasoning worse, not better.

**The exit 127 you hit is real and I'd like to fix it.** Same failure mode, different cause is the most likely story — `resolveClaudeCli()` returning nothing, or `getBinaryPathIfReady()` handing over a Mach-O `claude` that `launch.sh`'s Code-tab fix didn't reach. If you can get the actual argv, this reopens immediately with the right patch:

```
CLAUDE_COWORK_IPC_TAP=1 CLAUDE_COWORK_VERBOSE=1 ./launch.sh 2>&1 | tee ~/cowork.log
```

then any line matching `exec-capability` or `disclaimer`, plus your asar version from `./install.sh --doctor`. If `args[0]` really is `--` on your build, that's a bundle-shape difference worth recording in COMPAT.md, and this patch is the fix — I'd just want the version pinned to it.

---

johnzfitch:
For the record: the argv shape this PR guessed at has now been confirmed. #185 reports asar 1.40609.0 invoking the wrapper with `executableArgs: [ '--', '<claude path>' ]`, verbatim from the `[HostLoop] sdkOptions after patch` log, which is the evidence the closure above asked for. The same one-line skip landed via #186 (which carried that log and pinned the version) and is being consolidated in #188 with a COMPAT.md row. Your diagnosis of the failure mode was right; it just needed a build to point at.

---

PR #186, opened 2026-09-01
fix(exec-capability): unwrap the disclaimer `--` separator (#185)

### LINKED ISSUE #185: asar 1.40609.0: Cowork sessions exit 127, disclaimer unwrap reads the `--` separator as the command
## Summary

On asar **1.40609.0**, sending a message in a Cowork session fails immediately with the "Claude Code crashed" banner. `startup.log` reports `Claude Code process exited with code 127` a few milliseconds after spawn.

Area: Cowork session (spawn fails).

## Root cause

The bundle now invokes the disclaimer wrapper with a `--` separator:

```
[HostLoop] sdkOptions after patch: {
  executable: '/home/u/.config/Claude/Helpers/disclaimer',
  executableArgs: [ '--', '/home/u/.config/Claude/claude-code/2.1.247/claude.app/Contents/MacOS/claude' ],
  hasSpawnClaudeCodeProcess: true,
```

`resolveDisclaimerCommand()` reads the command from `args[0]`, so `cmd` is `"--"`. It matches neither the `claude.app` regex nor a `claude` basename, `realpathSafe("--")` returns null, and the unwrap gives up:

```
[exec-capability] BLOCKED (unresolvable): --
```

The interception in `frame-fix-wrapper.js` then passes the call through untouched, so the real `Helpers/disclaimer` stub runs, which is `exit 127` by design. This is #132's failure mode with a new argument shape.

Worth noting for anyone debugging this: the `Fix Code tab binary` block in `launch.sh` is not the lever here. It only matches the flat `claude-code/<ver>/claude` layout, while this build uses `claude-code/<ver>/claude.app/Contents/MacOS/claude`, and the bundle re-downloads the macOS binary at session start anyway, after the launcher has run. The unwrap substitution is what has to work.

## Steps to reproduce

1. Update to asar 1.40609.0
2. Open any conversation and send a message
3. The "Claude Code crashed" banner appears

## Relevant log output

```
[CCD] Installed bundle at /home/u/.config/Claude/claude-code/2.1.247
[exec-capability] BLOCKED (unresolvable): --
[HostLoop] cli.js stdin write failed (child likely exited): EPIPE
Session local_... query error: Claude Code process exited with code 127
  { errorClass: 'process_exited_nonzero', exitCode: 127 }
[QueryLoop] Exited ... after 5ms, exitPath=error: Claude Code process exited with code 127
```

## Environment

- CachyOS / KDE Plasma (Wayland)
- claude-cowork-linux `6c6ef96`
- Claude Desktop app.asar `1.40609.0`
- Electron `v42.1.0`, Node `v26.7.0`
- `install.sh --doctor`: 20 passed, 1 warning (asar newer than last tested), 0 failed

MouradFeddani:
Fixes #185.

## What does this change?

asar 1.40609.0 invokes the disclaimer wrapper as `disclaimer -- <cmd> [args...]`. `resolveDisclaimerCommand()` read the command from `args[0]`, so `cmd` was `"--"`, `realpathSafe` could not resolve it, and the unwrap returned null. The caller fell through to the exit-127 disclaimer stub instead of the resolved Claude binary, so every Cowork session died on spawn.

Skip a leading separator before reading cmd/rest. The older `<cmd> [args...]` shape keeps working.

## Type

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor

## Testing

- Distro / desktop: CachyOS / KDE Plasma (Wayland), Electron v42.1.0, Node v26.7.0, app.asar 1.40609.0
- Tested with `./install.sh --doctor`: 20 passed, 1 warning (asar newer than last tested), 0 failed
- Tested a full Cowork session: yes, sending a message in an existing conversation now spawns the CLI instead of returning the "Claude Code crashed" banner. `BLOCKED (unresolvable): --` and `exited with code 127` are both gone from `startup.log`.
- `tests/node/current-path/exec_capability_registry.test.cjs`: 41 passed, 0 failed. The added case fails on master and passes with the fix.
- `tests/test-cowork-patch.sh`: 93 passed, 0 failed.

## Security impact

- [x] This change does touch security-sensitive code

It sits in the disclaimer unwrap, which decides what gets spawned. The change is positional only: it selects which element of `args` is read as the command, and does not decide whether that command may run. Admission stays delegated to `resolveClaudeCli()` and `resolve()`, so the unwrap gains no policy of its own, per the comment above the function.

A path refused before is still refused behind the separator. The added test asserts both halves: `['--', <claude path>]` resolves to our binary with the separator consumed rather than forwarded, while `['--', '/opt/evil/hack', '--rm-rf']` and a bare `['--']` still return null.

## Checklist

- [x] `./install.sh --doctor` passes cleanly on my system

### DIFF
--- a/stubs/cowork/exec_capability_registry.js
+++ b/stubs/cowork/exec_capability_registry.js
@@ -246,8 +246,15 @@ function createExecCapabilityRegistry({
   // agrees with resolve() for every class is what keeps that true.
   function resolveDisclaimerCommand(args) {
     if (!Array.isArray(args) || args.length === 0) return null;
-    var cmd = args[0];
-    var rest = args.slice(1);
+    // Newer bundles (asar >= ~1.40609.0) invoke the wrapper as
+    // `disclaimer -- <cmd> [args...]`; older ones passed `<cmd> [args...]`.
+    // Without this, cmd is "--", realpath fails, and the caller falls through
+    // to the exit-127 disclaimer stub: #132's failure mode, new arg shape.
+    // Parsing only -- admission stays delegated to resolveClaudeCli()/resolve().
+    var argv = (args[0] === '--') ? args.slice(1) : args;
+    if (argv.length === 0) return null;
+    var cmd = argv[0];
+    var rest = argv.slice(1);
     // The asar invokes the Claude CLI through the disclaimer wrapper using
     // whatever path it chose -- a macOS-style claude.app/.../Claude path, or
     // the SDK path it installed (claude-code-vm/<ver>/claude), or a native
--- a/tests/node/current-path/exec_capability_registry.test.cjs
+++ b/tests/node/current-path/exec_capability_registry.test.cjs
@@ -251,6 +251,33 @@ describe('exec_capability_registry', () => {
         }
       });
 
+      // Regression for the asar >= ~1.40609.0 arg shape: the bundle now calls
+      // the wrapper as `disclaimer -- <cmd> [args...]`. Reading args[0] blindly
+      // made cmd "--", which realpath cannot resolve, so the unwrap returned
+      // null and the exit-127 stub ran: #132's failure mode with a new shape.
+      it('unwraps the `--` separator without widening admission', () => {
+        const reg = createExecCapabilityRegistry({
+          homedir: os.homedir(),
+          resolveClaudeBinaryPath: () => '/usr/local/bin/claude',
+        });
+
+        const result = reg.resolveDisclaimerCommand([
+          '--',
+          '/home/u/.config/Claude/claude-code/2.1.247/claude.app/Contents/MacOS/claude',
+          '-p', 'hi',
+        ]);
+        assert.ok(result, 'a separated call must unwrap, not fall through to the stub');
+        assert.strictEqual(result.cmd, '/usr/local/bin/claude');
+        assert.deepStrictEqual(result.rest, ['-p', 'hi'],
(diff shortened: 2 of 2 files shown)


johnzfitch:
Adopted in #188, which carries your commit unchanged with authorship intact, alongside #187 and two docs commits.

The evidence is what settles this one. #181 proposed the same skip in August and was closed because no call site in the in-tree bundle emitted a leading `--`; your `[HostLoop] sdkOptions after patch` log with `executableArgs: [ '--', '<claude path>' ]` is precisely the argv that closure asked for, and it pins the shape to 1.40609.0. The test is the right shape too: it fails on master (40/41), and the `['--', '/opt/evil/hack', '--rm-rf']` refusal is what makes "positional only, admission unchanged" a checked claim rather than a comment.

Two things carried into #188 on top of your commit: a `[PARTIAL]` COMPAT.md row for 1.40609.0 from your report, and an FAQ entry keyed on `BLOCKED (unresolvable): --` so the next person's log search lands on the fix. There is no pinning row, since the report did not record the CDN URL; if you still have the archive, its URL and `sha256sum` would complete it.

Leaving this open until #188 lands, so the fix stays visible on its own if that PR needs changes.

---

johnzfitch:
Landed: #188 merged to master with your commit carried unchanged and authorship intact. Closing this as adopted. Thanks for the fix and for the log that pinned the argv shape to a version.

---

PR #187, opened 2026-09-02
fix(bridge): fall back to ws when Electron has no net.WebSocket

leograba:
## Problem

The Claude-in-Chrome / remote-control bridge never connects on this port. The
Chrome extension is installed and Chrome is running, but the desktop app and the
extension never pair. In `startup.log`:

```
[bridge-ws] using net.WebSocket transport for wss://bridge.claudeusercontent.com/chrome/<uuid>
[claude-in-chrome] Failed to create WebSocket after 21ms: o.net.WebSocket is not a constructor
    at vJt (.../index.chunk-s0W1vGz6.js:729:74171)
    at Object.createWebSocket (.../index.chunk-s0W1vGz6.js:748:3950)
[claude-in-chrome] Bridge reconnecting in 30000ms (attempt 100)
[claude-in-chrome] Giving up bridge reconnection after 100 attempts
```

The bundle chooses between two transports behind a remote feature flag:

```js
async function yJt(e,t){
  if (t?.forceWs || !fJt()) { … return new _Jt(new ws(e, …)) }   // bundled `ws`
  return new gJt(vJt(e, t?.headers));                            // electron net.WebSocket
}
```

The flag is on, so it takes `net.WebSocket` — **which does not exist in the
Electron this port runs against.** Electron 42.1.0 exposes no `WebSocket` on the
`net` module at all (its `electron.d.ts` has a single, unrelated mention). The
constructor call throws synchronously, 21 ms in, before any network I/O.

**This misdiagnoses easily.** With the extension installed and the app signed in,
the in-app assistant reports it as "the extension isn't signed in / its session
with the Claude backend has lapsed" and sends you to re-authenticate in the side
panel. That can never fix it — there is no auth and no network involved in the
failure.

## Fix

A **capability test, not an override** — this is the part that matters for other
distros and other Electron builds:

- Where `net.WebSocket` **exists**, the condition is unchanged and the feature
  flag still decides. Behaviour is byte-identical to today for everyone whose
  Electron has the constructor, including any future Electron that adds it.
- Only builds that *would have thrown* fall through to the `ws` transport — the
  same path this code already takes whenever the flag is off or `forceWs` is set.
  It is not a new or untested code path.

`ws` is also the more capable transport: it is the only one implementing protocol
ping (`supportsProtocolPing: true`, vs the net transport whose `ping()` throws),
and the bundle already falls back to it at runtime when the peer requires ping.

Both decision sites are patched, and both are needed — the factory builds the
socket, while `wantedTransport()` reports which one is *wanted* and is
re-consulted by the `BF(flagId, …)` flag-change subscription. Patching only the
factory leaves the two disagreeing, with the subscription re-applying the flag.

Anchored on the `forceWs` / `forceWsTransport` property names, never the flag id
(`1284392461`) or the minified gate identifier — property names survive
minification, and the flag id is Anthropic's to rotate. `require()` is lazy, per
the existing note on the unguarded `resourcesPath` pass. Both guards match the
unpatched shape only, so re-runs are no-ops.

Added to `patch_index_apply_all` and nowhere else, per the ADDING A PASS contract,
so `launch.sh`, `PKGBUILD`/AUR and `install.sh` all pick it up.

## Verification

Against the **real** 1.30096.5 chunk, not only the fixture:

- Both sites rewrite correctly; `PATCH_INDEX_STRICT_SYNTAX=1` passes.
- Re-running `patch_index_apply_all` leaves the chunk byte-identical (idempotent).
- Patch present in the repacked `app.asar`, extracted back out and `node --check`ed.
- Runtime, current run of `startup.log` after relaunch:

```
[claude-in-chrome] WebSocket connected, sending connect message
[claude-in-chrome] Bridge received: {"type":"paired"}
[claude-in-chrome] Paired with Chrome extension (duration: 1194ms)
[claude-in-chrome] ensureConnected called, connected=true, authenticated=true, wsState=1
[claude-in-chrome] Tool call completed: tabs_context_mcp (f8eaad39) in 769ms
```

`using net.WebSocket transport`, `not a constructor` and `Giving up bridge
reconnection` are all **absent** from that run. End-to-end exercised through the
real browser: navigation, DOM/JS access, console reading and network capture.

`tests/test-cowork-patch.sh`: **99 passed, 0 failed** (6 new assertions — both
sites, on both the double-quoted and backtick fixtures, plus refutations of the
unpatched shapes). The pass-count guard is bumped 9 → 11.

## Environment

| | |
|:--|:--|
| Distro | Ubuntu 24.04.4 LTS, kernel 7.0.0-30-generic x86_64 |
| Electron | **42.1.0** (npm, `~/.local/lib/node_modules/electron`) — the affected variable |
| Claude Desktop asar | 1.30096.5 |
| Node (system) | v18.19.1 (`/usr/bin/node`) |
| Node (nvm, used for install.sh) | v24.11.1 — nvm also has v18.10.0, v20.12.2, v20.15.0 |
| npm | 11.6.2 |
| Google Chrome | 152.0.7977.64, extension `fcoeoabgfenejglbffodgkkbkcdhcgfn` v1.0.90 |
| Install | `install.sh` → `~/.local/share/claude-desktop` |

`install.sh` needs Node ≥ 22 on PATH (nvm v24 here): `asar.mjs` uses
`import … with { type: "json" }`, which the system v18.19.1 rejects.

## Distro / packaging impact

- **No behaviour change** on any Electron that has `net.WebSocket` — the flag
  still decides there.
- **No new dependency**: `ws` is already bundled and already used by this code.
- **AUR/`PKGBUILD`**: picked up automatically via the shared pass list; runs under
  `PATCH_INDEX_STRICT_SYNTAX=1`, which passed on the real chunk.
- **Older/newer asars**: grep-guarded like every other pass, so a build without
  these shapes is a clean no-op (it will emit the usual WARN-on-miss under
  `PKGBUILD`, and is silenced in the launchers as with the other passes).
- Electron 42.1.0 is what this port installs today, so the affected path is the
  default for anyone following the README, not an exotic setup.

### DIFF
--- a/patch-index.sh
+++ b/patch-index.sh
@@ -257,6 +257,53 @@ patch_index_apply_all() {
     'process\.resourcesPath,["`]app\.asar["`]' \
     's/process\.resourcesPath,["`]app\.asar["`]/require("electron").app.getAppPath()/g'
 
+  # Fix the Claude-in-Chrome / remote-control bridge transport on Electron builds
+  # that have no net.WebSocket.
+  #
+  # The bridge picks between Electron's net.WebSocket and the bundled `ws`
+  # package, behind a remote feature flag. With the flag on it takes the
+  # net.WebSocket path -- but electron.net.WebSocket does not exist in every
+  # Electron (Electron 42.1.0 exposes no WebSocket on the net module at all), and
+  # the constructor call then throws synchronously:
+  #
+  #   [bridge-ws] using net.WebSocket transport for wss://bridge.claudeusercontent.com/chrome/<uuid>
+  #   [claude-in-chrome] Failed to create WebSocket after 21ms: o.net.WebSocket is not a constructor
+  #   [claude-in-chrome] Giving up bridge reconnection after 100 attempts
+  #
+  # The visible symptom is a Chrome extension that never pairs with the desktop
+  # app, which reads as an extension or sign-in problem. It is neither: the
+  # failure is 21ms in, before any network I/O.
+  #
+  # This is a CAPABILITY TEST, not an override. Where net.WebSocket exists the
+  # condition is unchanged and the flag still decides, so builds on a newer or
+  # patched Electron behave exactly as upstream intends; only builds that would
+  # have thrown fall through to the `ws` transport the same code already uses
+  # whenever the flag is off or forceWs is set. That path is also the more
+  # capable one -- it is the only transport implementing protocol ping, and the
+  # bundle already falls back to it at runtime when the peer requires ping.
+  #
+  # Two sites, and both are needed: the factory decides which socket is built,
+  # while wantedTransport() reports which one is wanted and is re-consulted by
+  # the flag-change subscription. Patching only the factory leaves the two
+  # disagreeing, and the subscription re-applies the flag on top.
+  #
+  # Anchored on the forceWs / forceWsTransport property names, never on the flag
+  # id or the minified gate identifier: property names survive minification, and
+  # the flag id is Anthropic's to rotate. require() lazily, per the note on the
+  # unguarded resourcesPath pass -- these are CJS chunks and no minified electron
+  # binding is guaranteed to be in scope here.
+  #
+  # Both guards match the unpatched shape only, so re-runs are no-ops: the
+  # factory guard requires the condition to be immediately followed by `){`,
+  # and the wantedTransport guard requires `<ident>()` directly before the
+  # `&&!this.forceWsTransport`. After patching neither holds.
+  patch_index "Patching bridge transport to fall back to ws when net.WebSocket is absent..." \
+    '\?\.forceWs\|\|![A-Za-z0-9_$]+\(\)\)\{' \
+    's/(\?\.forceWs\|\|![A-Za-z0-9_$]+\(\))\)\{/\1||typeof require("electron").net.WebSocket!="function"){/g'
+  patch_index "Patching bridge wantedTransport() to match the net.WebSocket capability..." \
+    '[A-Za-z0-9_$]+\(\)&&!this\.forceWsTransport' \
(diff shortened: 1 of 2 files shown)


leograba:
p.s. Claude created this solution entirely by itself, I am not knowledgeable to understand how good it is for merging upstream. My intent is to at least inform about the issue and share what Claude came up with as a workaround

johnzfitch:
Adopted in #188, which carries your commit unchanged with authorship intact, alongside #186 and two docs commits.

On your note that you could not judge it for upstream: I read it against the pass contract and it holds up. The gate is a capability test rather than an override, so where `net.WebSocket` exists the flag still decides and nothing changes; patching both the factory and `wantedTransport()` is necessary, because the flag-change subscription re-consults the latter and would otherwise re-apply the flag over the factory's choice; both guards match only the unpatched shape, so the passes are idempotent (the existing re-run test covers the new fixture lines); and `require("electron")` has precedent in the unguarded resourcesPath pass. The 99/99 suite result reproduces here, in isolation on your branch and after the cherry-pick, and `PATCH_INDEX_STRICT_SYNTAX=1` is clean on the backtick fixture. What I cannot do from this environment is run it against a real 1.30096.5 chunk, so your verification against the live bundle is the evidence for that half.

Carried into #188 on top of your commit: a `[PARTIAL]` COMPAT.md row for 1.30096.5, an FAQ entry keyed on `o.net.WebSocket is not a constructor` (with the misdiagnosis you flagged called out), and a note in extensions.md that newer bundles pair over the bridge rather than the native host that section describes. No pinning row, since the report did not record the CDN URL.

Leaving this open until #188 lands.

---

johnzfitch:
Landed: #188 merged to master with your commit carried unchanged and authorship intact. Closing this as adopted. Thanks for the fix and for verifying it against the real chunk.

---

leograba:
Nice, I'm glad to have helped :) have a nice day

